PR #5, opened 2019-02-05
Ce 569/peer group skeleton

jtownson:
First story towards implementation described at https://input-output.atlassian.net/wiki/spaces/CE/pages/517963843/CE-566+-+Peer+Group+Foundations.

* Removes all legacy code and dependencies.
* Add the basic peer group abstraction.
* Provide users choice of Future/EitherT/IO/Task.
* Impl of TCPPeerGroup copy/pasted from CE's NettyTransport (now with tests:-).


src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -0,0 +1,24 @@
...
+  def sendMessage(address: A, message: ByteBuffer): F[Unit]
+  def shutdown(): F[Unit]
+  def messageStream(): MessageStream[ByteBuffer]
+}
+
+object PeerGroup {
+
+  type Lift[F[_]] = Kleisli[F, Task[Unit], Unit]
+
+  abstract class TerminalPeerGroup[A, F[_]] extends PeerGroup[A, F]
```
AlexITC: Why an abstract class here and a trait for `PeerGroup`? we should be consistent.

src/io/iohk/scalanet/peergroup/Subscribers.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.util.concurrent.CopyOnWriteArraySet
+
+import monix.reactive.{Observable, OverflowStrategy}
+import monix.reactive.observers.Subscriber
+
+import scala.collection.mutable
+import scala.collection.JavaConverters._
+
+private[scalanet] class Subscribers {
+  val subscriberSet: mutable.Set[Subscriber.Sync[ByteBuffer]] =
```
AlexITC: Can we make this private and create a method to update the set, this would avoid leaking implementation details.

src/io/iohk/scalanet/peergroup/Subscribers.scala
```diff
@@ -0,0 +1,25 @@
...
+    new CopyOnWriteArraySet[Subscriber.Sync[ByteBuffer]]().asScala
+
+  val monixMessageStream: Observable[ByteBuffer] =
```
AlexITC, line 16: The braces aren't expensive and can avoid us issues on methods like this one.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  private val clientBootstrap = new Bootstrap()
+    .group(workerGroup)
+    .channel(classOf[NioSocketChannel])
+    .option[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, true)
+
+  private val serverBootstrap = new ServerBootstrap()
+    .group(workerGroup)
+    .channel(classOf[NioServerSocketChannel])
+    .childHandler(new ChannelInitializer[SocketChannel]() {
```
AlexITC, line 38: Looks like we can use the single-abstract method syntactic sugar.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -0,0 +1,118 @@
...
+    .option[Integer](ChannelOption.SO_BACKLOG, 128)
+    .childOption[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, true)
+    .bind(config.bindAddress)
+    .syncUninterruptibly()
+
+  override def sendMessage(address: InetSocketAddress, message: ByteBuffer): F[Unit] = {
+    val send: Task[Unit] = Task {
+
+      val activationAdapter = new ChannelInboundHandlerAdapter() {
+        override def channelActive(ctx: ChannelHandlerContext): Unit = {
```
AlexITC, line 55: Looks like we can use the single-abstract method syntactic sugar.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -0,0 +1,118 @@
...
+          ctx
+            .writeAndFlush(Unpooled.wrappedBuffer(message))
+            .addListener((_: ChannelFuture) => ctx.channel().close())
+        }
+      }
+
+      clientBootstrap
+        .handler(new ChannelInitializer[SocketChannel]() {
+          def initChannel(ch: SocketChannel): Unit = {
```
AlexITC, line 64: Looks like we can use the single-abstract method syntactic sugar.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  override def shutdown(): F[Unit] = {
+    liftF(Task {
+      serverBootstrap.channel().close()
+      workerGroup.shutdownGracefully()
+      ()
+    })
+  }
+
+  private val subscribers = new Subscribers()
```
AlexITC: Class attributes are usually placed at the beginning of the class to make readers aware of them.

src/io/iohk/scalanet/test/NetUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+      new InetSocketAddress("localhost", s.getLocalPort)
+    } finally {
+      s.close()
+    }
+  }
+
+  def isListening(address: InetSocketAddress): Boolean = {
+    try {
+      new Socket(address.getHostName, address.getPort).close()
+      true
```
AlexITC, line 22: If I understand correctly, the address is used (or listening) if we aren't able to create the socket, which means that we must return false.

AlexITC (commented):
I would like to see a test that checks that a huge message can be sent using the library, if this is not adequate for this task, please create a ticket.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
shaileshp0110: This trait has more than 1 method

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
shaileshp0110: same as above

src/io/iohk/scalanet/test/NetUtils.scala
shaileshp0110: This is client socket connection to the server.If it succeeds then server is listening

AlexITC (approved):

PR #6, opened 2019-02-08
Ce 570/create udp terminal

jtownson:
Adds basic UDP support.


src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -21,4 +21,24 @@ object PeerGroup {
...
+  def createOrThrow[PG](pg: => PG, config: Any): PG =
+    try {
+      pg
+    } catch {
+      case t: Throwable =>
+        throw new IllegalStateException(initializationErrorMsg(config), t)
+    }
+
+  private def initializationErrorMsg[F[_]](config: Any) =
+    s"Failed initialization of ${classOf[TCPPeerGroup[F]].getName} with config $config. Cause follows."
```
AlexITC, line 42: Have you tried the usefulness of printing `Config`? Sometimes it can be just too noisy without the actual details needed to debug, it could print a useless string too.

Also, the `classOf[TCPPeerGroup[F]].getName` is not accurate anymore.

src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -21,4 +21,24 @@ object PeerGroup {
   trait TerminalPeerGroup[A, F[_]] extends PeerGroup[A, F]
 
   case class InitializationError(message: String, cause: Throwable)
+
+  def create[PG](pg: => PG, config: Any): Either[InitializationError, PG] =
```
AlexITC, line 25: I wonder if it would be better to receive `pg: => Try[PG]` instead to make it obvious that an exception can be thrown.

Why is the config required if the peer group can be created without it? If you keep it, please add a type parameter for the config, `Any` is just too general.

Another option is to update the method to `def create[PG, C](pg: C => PG, config: C)`

AlexITC (commented):
A few minor comments.

PR #10, opened 2019-02-22
CE-596 Use new bazel deps on master

marcesquerra:


AlexITC (approved):

src/io/iohk/network/DestinationDescriptor.scala
```diff
@@ -0,0 +1,29 @@
+package io.iohk.network
```
vargasmontero, line 1: Why is this file in this PR?

vargasmontero (approved):

src/io/iohk/network/DestinationDescriptor.scala
marcesquerra: CEF should be depending on commits from master of `scalanet`. But it's currently depending on a commit from `develop`. Since the purpose of this PR is to upgrade whatever CEF is using to use the new bazel-deps, I've merged whatever it was missing too so that master has all CEF needs.

PR #13, opened 2019-03-04
Ce 601/pg interfaces/f

jtownson:
Exposing monix from the interface. Deletion of the 'F' abstraction.


src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
```diff
@@ -59,43 +60,41 @@ class SimplePeerGroup[A, F[_], AA](
...
-          .head()
-      }
-
-      liftF(enrolledTask)
+      val enrolledTask: Task[Unit] = controlChannel.inboundMessages.collect {
+        case Enrolled(_, _, newRoutingTable) =>
+          routingTable.clear()
+          routingTable ++= newRoutingTable
+          log.debug(s"Peer address '$processAddress' enrolled into group and installed new routing table:")
+          log.debug(s"$newRoutingTable")
```
AlexITC: Please, merge the logs to find these details easily on concurrent environments.

AlexITC (approved):

src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
jtownson: Done. I guess in fact correlation ids are the only way to make sense of logs in concurrent environments.

PR #25, opened 2019-06-01
A 574/midnight/californium dtls

jtownson:
This is the DTLS part of the encryption story.


src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+
+  override def initialize(): Task[Unit] = {
+    Task(serverConnector.start())
+  }
+
+  override def client(to: InetMultiAddress): Task[Channel[InetMultiAddress, M]] = {
+    val connector = createClientConnector()
+    connector.start()
+    val id = (connector.getAddress, to.inetSocketAddress)
+    assert(!activeChannels.contains(id), s"HOUSTON, WE HAVE A MULTIPLEXING PROBLEM")
```
KonradStaniec, line 44: from the perspective of he user of this library, it would be nice to now when this condition can happen, maybe some documentation or something.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+
+        override def onContextEstablished(endpointContext: EndpointContext): Unit = ()
+
+        override def onSent(): Unit = promisedSend.success(())
+
+        override def onError(throwable: Throwable): Unit = promisedSend.failure(throwable)
+      }
+
+      val rawData = RawData.outbound(buffer.toArray, new AddressEndpointContext(to.inetSocketAddress), callback, false)
+      dtlsConnector.send(rawData)
```
KonradStaniec: what if encoded data will be bigger than maximum datagram size, will it throw some exception ?

src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object InetPeerGroupUtils {
+  def toTask(f: util.concurrent.Future[_]): Task[Unit] = {
+    val promisedCompletion = Promise[Unit]()
+    f.addListener((_: util.concurrent.Future[_]) => promisedCompletion.complete(Success(())))
+    Task.fromFuture(promisedCompletion.future)
+  }
+
+  def getChannelId(remoteAddress: InetSocketAddress, localAddress: InetSocketAddress): Seq[Byte] = {
+    val b = ByteBuffer.allocate(16)
```
KonradStaniec, line 28: Will it not fail for ip6 ?

dependencies.yaml
```diff
@@ -40,5 +41,12 @@ dependencies:
       version: "2.24.0"
       lang: java
       scope: test
+  org.eclipse.californium:
```
KonradStaniec, line 44: hmm it seems califronium is `Eclipse Californium is a Java implementation of RFC7252 - Constrained Application Protocol for IoT Cloud services` is library which focuses on specific goals and need. Wouldn't it be better to use generic cryptography library which supports both tls and dtls, like https://bouncycastle.org/ ?

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+    TLS_ECDHE_ECDSA_WITH_AES_256_CCM,
+    TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
+    TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,
+    TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384
+  )
+
+  trait Config {
+    val bindAddress: InetSocketAddress
+    val processAddress: InetMultiAddress
+    private[scalanet] def scandiumConfigBuilder: DtlsConnectorConfig.Builder
```
KonradStaniec, line 171: there are a lot of setting in this `DtlsConnectorConfig.Builder` , it would be nice to have at least documentation of all of them, or maybe case class which captures all of them with some defaula. As a user of library it would great to be able configure most of connection properties, without need to change library code.

src/io/iohk/scalanet/test/peergroup/DTLSPeerGroupSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+import DTLSPeerGroupSpec._
+import io.iohk.scalanet.peergroup.DTLSPeerGroup.Config
+
+class DTLSPeerGroupSpec extends FlatSpec {
+
+  implicit val patienceConfig = PatienceConfig(5 seconds)
+
+  behavior of "DTLSPeerGroup"
+
+  it should "send and receive a message" in {
```
KonradStaniec, line 33: i it would probably be nice to have some negative test cases like:
what if there is not server listening on address that client tries to connect ?
what if one of peer disconect during peer dtls handshake ? 
what if one peer will have wrong keys and certifiicates ? 
what if peer won't support any of the ciphers ?

KonradStaniec (commented):
as a general note i think more documentation would be nice to have, or examples how to use some of the functions and primitives. I mean i can always look into tests to check it out, but it would be troublesome to do it everytime.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
jtownson: Okay, okay, documentation is crap :-) Will add some.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
jtownson: Good question. Will investigate.

There is a story for error handling btw.

src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
jtownson: Yep. We have a story to fix.

dependencies.yaml
jtownson: As far as I know, the scandium subproject contains a complete implementation of DTLS and the code quality is certainly very good. Nonetheless, I wish I had been aware of bouncy castle's DTLS support. I will investigate in the background.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
jtownson: Yeah, of course there are loads of configuration options that a production implementation might need. I plan to drive the addition of all of these from real use-cases.

What I am keen to avoid is spending a lot of time developing a full-featured, error free, DTLS implementation and then find we don't use it, or we only ever use it in one configuration, etc.

src/io/iohk/scalanet/test/peergroup/DTLSPeerGroupSpec.scala
jtownson: Sure. A lot of this stuff will come in the future if/when we decide to use DTLS (e.g. for discovery).

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+  override def shutdown(): Task[Unit] =
+    for {
+      _ <- Task(serverConnector.stop())
+      _ <- Task(serverConnector.destroy())
+    } yield ()
+
+  private class ChannelImpl(val to: InetMultiAddress, dtlsConnector: DTLSConnector)(implicit codec: Codec[M])
+      extends Channel[InetMultiAddress, M] {
+
+    override val in: Subject[M, M] = ReplaySubject[M]()
```
AlexITC, line 61: Are you aware that anyone consuming this stream will get all events from the beginning instead of the ones after subscribing, right? otherwise, it can cause weird behavior.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
AlexITC: A simple way is to let the user provide the `DtlsConnectorConfig` or the builder.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+      certificate validation requires that root keys be distributed
+      independently, the self-signed certificate that specifies the root
+      certificate authority MAY be omitted from the chain, under the
+      assumption that the remote end must already possess it in order to
+      validate it in any case.
+     */
+    case class CertAuthenticated(
+        bindAddress: InetSocketAddress,
+        processAddress: InetMultiAddress,
+        certificateChain: Array[Certificate],
```
AlexITC, line 218: I suggest to avoid exposing arrays but safe collections like `List[T]` instead.

src/io/iohk/scalanet/test-resources/create-keystore.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash +x
```
AlexITC, line 1: +1

AlexITC (commented):
A-574 is supposed to implement TLS instead of DTLS, if the goal changed, please update the task accordingly.

Besides that, it looks good to me, please create tickets to address @KonradStaniec comments.

src/io/iohk/scalanet/test-resources/create-keystore.sh
loverdos: `/usr/bin/env bash` is usually the preferred approach. `/bin/bash` does not exist everywhere.

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+
+  private val activeChannels = new ConcurrentHashMap[(InetSocketAddress, InetSocketAddress), ChannelImpl]().asScala
+
+  override def processAddress: InetMultiAddress = config.processAddress
+
+  override def initialize(): Task[Unit] = {
+    Task(serverConnector.start())
+  }
+
+  override def client(to: InetMultiAddress): Task[Channel[InetMultiAddress, M]] = {
```
krcz: return type (`Task`) suggests that creation is going to be done asynchronously (and lazily), while implementation is synchronous, returning constant Task. moreover all errors are going to be thrown instead of encapsulated in `Task` (as one might expect).

do you think it would be possible to modify implementation to make it more idiomatic?

src/io/iohk/scalanet/peergroup/DTLSPeerGroup.scala
```diff
@@ -0,0 +1,242 @@
...
+  private class ChannelImpl(val to: InetMultiAddress, dtlsConnector: DTLSConnector)(implicit codec: Codec[M])
+      extends Channel[InetMultiAddress, M] {
+
+    override val in: Subject[M, M] = ReplaySubject[M]()
+
+    override def sendMessage(message: M): Task[Unit] = {
+      import io.iohk.scalanet.peergroup.BufferConversionOps._
+      val buffer = codec.encode(message)
+
+      val promisedSend = Promise[Unit]()
```
krcz: wouldn't `Task.create` be better here? especially that you are creating side effect by sending message.

src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import io.netty.util
+import monix.eval.Task
+
+import scala.concurrent.Promise
+import scala.util.Success
+
+object InetPeerGroupUtils {
+  def toTask(f: util.concurrent.Future[_]): Task[Unit] = {
+    val promisedCompletion = Promise[Unit]()
```
krcz, line 22: could you try something like:
```
    Task.create { (_, cb) =>
      f.addListener { (_: util.concurrent.Future[_]) =>
        if (f.isSuccess()) cb.onSuccess(f.get()) else cb.onError(f.cause()) 
      }
    }
```

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -33,7 +36,8 @@ import scala.util.Success
   * @param codec a decco codec for reading writing messages to NIO ByteBuffer.
   * @tparam M the message type.
   */
-class TCPPeerGroup[M](val config: Config)(implicit codec: Codec[M]) extends TerminalPeerGroup[InetMultiAddress, M]() {
+class TCPPeerGroup[M](val config: Config)(implicit codec: Codec[M], bufferInstantiator: BufferInstantiator[ByteBuffer])
```
krcz, line 39: I don't know what `BufferInstantiator` is, so that might be obvious, but is there any reason you're making it configurable in that PR?

krcz (commented):
I have some questions / suggestions regarding `Task` usage

jtownson:
Closing. Merged via command line merge --squash.

PR #27, opened 2019-06-04
A 574/tls peer group

shaileshp0110:


src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+    for {
+      _ <- toTask(serverBind.channel().close())
+      _ <- toTask(workerGroup.shutdownGracefully())
+    } yield ()
+  }
+}
+
+object TLSPeerGroup {
+
+  case class Config(
```
jtownson, line 106: Please can you make this identical to the config in DTLSPeerGroup? Either by copying/pasting the config from there or by refactoring it into some common location.

src/io/iohk/scalanet/test/peergroup/TLSPeerGroupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.Random
+
+class TLSPeerGroupSpec extends FlatSpec with BeforeAndAfterAll {
+
+  implicit val patienceConfig: ScalaFutures.PatienceConfig = PatienceConfig(5 seconds)
+
+  behavior of "TCPPeerGroup"
```
jtownson: copy+paste error

src/io/iohk/scalanet/test/peergroup/TLSPeerGroupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+  it should "send and receive a message when client auth is disabled / false" in
+    withTwoRandomTLSPeerGroups[String](false) { (alice, bob) =>
+      println(s"Alice is ${alice.processAddress}, bob is ${bob.processAddress}")
+      val alicesMessage = Random.alphanumeric.take(1024).mkString
+      val bobsMessage = Random.alphanumeric.take(1024).mkString
+
+      bob.server().foreachL(channel => channel.sendMessage(bobsMessage).evaluated).runAsync
+      val bobReceived: Future[String] = bob.server().mergeMap(channel => channel.in).headL.runAsync
+      val aliceClient = alice.client(bob.processAddress).evaluated
+      Thread.sleep(2000)
```
jtownson: needs to work without this sleep

src/io/iohk/scalanet/test/peergroup/TLSPeerGroupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.util.Random
+
+class TLSPeerGroupSpec extends FlatSpec with BeforeAndAfterAll {
+
+  implicit val patienceConfig: ScalaFutures.PatienceConfig = PatienceConfig(5 seconds)
+
+  behavior of "TCPPeerGroup"
+
+  it should "send and receive a message when client auth is disabled / false" in
+    withTwoRandomTLSPeerGroups[String](false) { (alice, bob) =>
```
jtownson: please can you test using both an authenticated and unauthenticated configuration (there is code you can borrow from DTLSPeerGroupSpec for that).

src/io/iohk/scalanet/test/peergroup/TLSPeerGroupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+      val aliceClient = alice.client(bob.processAddress).evaluated
+      Thread.sleep(2000)
+      val aliceReceived = aliceClient.in.headL.runAsync
+      aliceClient.sendMessage(alicesMessage).evaluated
+
+      bobReceived.futureValue shouldBe alicesMessage
+      aliceReceived.futureValue shouldBe bobsMessage
+    }
+
+  it should "send and receive a message when client auth is enabled or true" in
```
jtownson: See comment above about refactoring into a single test with two configs, one for authenticated one for unauthenticated.

jtownson (changes requested):
Please don't merge this into develop. Merge down into the A-574 branch so we can pull together TLS and DTLS into one place.

.gitignore
```diff
@@ -7,3 +7,5 @@ target/
 bazel-*
 .bazel-*
 .dazel_run
+
+scalafmt
```
AlexITC, line 11: why?

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+
+    private val sslServerCtx = SslContextBuilder.forServer(config.certChainPrivateKey, config.certChain: _*).build()
+
+    log.debug(
+      s"Creating server channel from ${nettyChannel.localAddress()} to ${nettyChannel.remoteAddress()} with channel id ${nettyChannel.id}"
+    )
+
+    nettyChannel
+      .pipeline()
+      .addLast("ssl", sslServerCtx.newHandler(nettyChannel.alloc())) //This needs to first
```
AlexITC: to **be** first

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+      .addLast("frameDecoder", new LengthFieldBasedFrameDecoder(Int.MaxValue, 0, 4, 0, 4))
+      .addLast("frameEncoder", new LengthFieldPrepender(4))
+      .addLast(new MessageNotifier(messageSubject))
+
+    override val to: InetMultiAddress = InetMultiAddress(nettyChannel.remoteAddress())
+
+    override def sendMessage(message: M): Task[Unit] = {
+      toTask({
```
AlexITC: Are the braces required?

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+class TLSPeerGroup[M](val config: Config)(implicit codec: Codec[M]) extends TerminalPeerGroup[InetMultiAddress, M]() {
+
+  private val log = LoggerFactory.getLogger(getClass)
+
+  private[scalanet] class ServerChannelImpl[M](val nettyChannel: SocketChannel)(
+      implicit codec: Codec[M]
+  ) extends Channel[InetMultiAddress, M] {
+
+    private val log = LoggerFactory.getLogger(getClass)
+    private val messageSubject = ReplaySubject[M]()
```
AlexITC: Be aware that anyone subscribing to this subject will receive the whole message history instead of the ones produced after subscribing to the subject, if that's the intended behavior, please add a comment explaining why.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+      .build()
+
+    private val bootstrap: Bootstrap = clientBootstrap
+      .clone()
+      .handler(new ChannelInitializer[SocketChannel]() {
+        def initChannel(ch: SocketChannel): Unit = {
+          val pipeline = ch.pipeline()
+          val sslHandler = sslClientCtx.newHandler(ch.alloc())
+          pipeline
+            .addLast("ssl", sslHandler) //This needs to first
```
AlexITC: to **be** first

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+            .addLast("frameEncoder", new LengthFieldPrepender(4))
```
AlexITC: These lines are equivalent to what is being done on lines around `51`, create some constants and reuse them here.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+            .addLast(new ChannelInboundHandlerAdapter() {
+              override def channelInactive(ctx: ChannelHandlerContext): Unit = {
+                deactivation.success(())
+              }
+
+              override def userEventTriggered(ctx: ChannelHandlerContext, evt: Any): Unit = {
+                evt match {
+                  case e: SslHandshakeCompletionEvent =>
+                    if (e.isSuccess) activation.success(ctx) else activation.failure(e.cause())
+                    log.debug(
```
AlexITC: Possibly safer to log before the if.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+    private val messageSubject = ReplaySubject[M]()
+
+    private val sslClientCtx = SslContextBuilder
+      .forClient()
+      .trustManager(config.trustStore: _*)
+      .build()
+
+    private val bootstrap: Bootstrap = clientBootstrap
+      .clone()
+      .handler(new ChannelInitializer[SocketChannel]() {
```
AlexITC: What about removing unnecessary nesting? There is more nesting while adding the `ChannelInboundHandlerAdapter` too.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  private val channelSubject = PublishSubject[Channel[InetMultiAddress, M]]()
+
+  private val workerGroup = new NioEventLoopGroup()
+
+  private val clientBootstrap = new Bootstrap()
+    .group(workerGroup)
+    .channel(classOf[NioSocketChannel])
+    .option[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, true)
```
AlexITC, line 64: These values look like good candidates to be configurable from scalanet users.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+    .option[Integer](ChannelOption.CONNECT_TIMEOUT_MILLIS, 10000)
+
+  private val serverBootstrap = new ServerBootstrap()
+    .group(workerGroup)
+    .channel(classOf[NioServerSocketChannel])
+    .childHandler(new ChannelInitializer[SocketChannel]() {
+      override def initChannel(ch: SocketChannel): Unit = {
+        val newChannel = new ServerChannelImpl[M](ch)
+        channelSubject.onNext(newChannel)
+        log.debug(s"$processAddress received inbound from ${ch.remoteAddress()}.")
```
AlexITC, line 75: It can be a good idea to log before feeding the subject, one advantage would be to simplify tracing, receiving the item and seeing the log after that could cause confusion.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -0,0 +1,298 @@
...
+      }
+    })
+    .option[Integer](ChannelOption.SO_BACKLOG, 128)
+    .option[RecvByteBufAllocator](ChannelOption.RCVBUF_ALLOCATOR, new DefaultMaxBytesRecvByteBufAllocator)
+    .childOption[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, true)
```
AlexITC, line 80: Also, nice to have a way to set them from the outside.

AlexITC (commented):
I'm concerned that this PR contains Jeremy's work from DTLS, it makes it very difficult to review your actual changes, besides that, I left some suggestions.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
KonradStaniec: It would be probably good idea to enforce `TLS 1.2` for both client and server 
```
val handler = sslContext.newHandler(socketChannel.alloc());
handler.engine().setEnabledProtocols(new String[] {"TLSv1.2"});
```
Another option worth considering, is to configure cipher suites which maintain perfect forward secrecy, in TLS those are cipher suites based on DH and ECDHE key exchanges.

It is worth noting that form TLS 1.3 only such cipher suites will be available, but unfortunatly TLS 1.3 is only available for java 11 which do not play good with scala yet.

KonradStaniec: It would be probably good idea to enforce `TLS 1.2` for both client and server 
```
val handler = sslContext.newHandler(socketChannel.alloc());
handler.engine().setEnabledProtocols(new String[] {"TLSv1.2"});
```
Another option worth considering, is to configure cipher suites which maintain perfect forward secrecy, in TLS those are cipher suites based on DH and ECDHE key exchanges.

It is worth noting that form TLS 1.3 only such cipher suites will be available, but unfortunatly TLS 1.3 is only available for java 11 which do not play good with scala yet.

shaileshp0110: JDK 8 will use TLS 1.2 as default
https://docs.oracle.com/javase/8/docs/technotes/guides/security/enhancements-8.html
https://blogs.oracle.com/java-platform-group/jdk-8-will-use-tls-12-as-default

.gitignore
jtownson: I've changed the script to install scalafmt locally so that it does not require sudo privs. This stops you accidentally adding a 17Mb jar to the project :-)

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
jtownson: This an ongoing issue. We had a test failure (about a month ago) where the new channel event was not delivered to subscribers until _after_ the first message had been delivered to that channel's message stream. This meant that the subscriber missed the start of the message stream.

The only suggestion to avoid this problem was to replay the message stream.

Nonetheless, this replay approach is potentially problematic is production scenarios because of the obvious memory problems associated with replay buffering. We need a confirmed and robust solution that can be rolled out to peer group implementations. Obviously we don't want to block scalanet development in the meantime so replaying is there for the time being.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
jtownson: Sure, but we will only drive this stuff from use cases. If somebody needs the option, we add it. Otherwise we hard code it.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
jtownson: Agreed re ECDHE ciphers

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
AlexITC: Ok, please create a ticket, otherwise, we may forget.

jtownson:
Closing. Merged via command line merge --squash.

PR #29, opened 2019-06-12
Question related to server method

EzequielPostan:
Test added to discuss behaviour

### DIFF
--- a/src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
+++ b/src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
@@ -8,7 +8,6 @@ import scala.concurrent.Future
 import scala.concurrent.duration._
 import io.iohk.decco.auto._
 import monix.execution.Scheduler.Implicits.global
-
 import org.scalatest.concurrent.ScalaFutures._
 import io.iohk.scalanet.TaskValues._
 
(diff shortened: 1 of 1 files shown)


krcz:
That's a very good question! In case of TCP protocol itself has a notion of connection, while UDP doesn't. I see two choices: one is to implement it on our side, the other is to have both connection-aware and connectionless peer groups - in the first one connection could be present with no messages incoming, in the second it starts to exist with the first message incoming.

But there is dual problem as well: if the protocol doesn't have the notion of connection it means that it cannot be closed either. If we plan having peer groups with dynamic membership - with unbound number of members - that can lead to resources leak.

AlexITC:
> if the protocol doesn't have the notion of connection it means that it cannot be closed either

Exactly, if we want to simulate a connection on UDP, we'd need to simulate disconnection events, which is likely to require a timer that you reset each time a message is received, when it expires, it fires the disconnection event.

You can look into how DTLS handles this, I bet they use the timer approach.

The important question is, do we really need to simulate a connection?

EzequielPostan:
Some conclusions from a recent call with @shaileshp0110  
Things depend on how you implement channels and the use case you need.
In my case, if the channel will "save" the messages I receive until I subscribe to the stream of messages the channel provides to me, then no, you don't necessarily need the connection. If the messages are not saved, then you need to know about connections in order to avoid the loss of messages. Currently, messages are saved until the receiver calls the `ìn` method, so I am good without connections. 
I haven't thought about other use cases. On a similar note, TCPPeerGroup should behave in the same way as this one, i.e. the channel is not received until the first message is sent (which is a separate connection semantic from the underlying TCP connection).

jtownson:
It is an interesting case. I think the answer is as follows. 
* There is no need/requirement for `bob.client.evaluated` in your example to eagerly create a corresponding channel notification with `alice.server`.
* The requirement is looser: a new channel event must be received on the server _before_ the first message.

I recommend we think less about channels and messaging and more about 'synchronising state' between two peers (or threads/processes within peers, according to how well multiplexing works). A peer has some state, which we call a message, and it wants another peer to have that state also. If it can achieve this goal then all is good. This is the rationale by which we should evaluate such use-cases.

EzequielPostan:
> The requirement is looser: a new channel event must be received on the server before the first message.

That clarifies this question. Hence I will proceed to close this PR.
I will probably open a new one soon with further questions.
Thank you @krcz @AlexITC and @jtownson for the interest in the topic.

PR #30, opened 2019-06-13
added the ipv4 and ipv6 test for getChannelId

shaileshp0110:


src/io/iohk/scalanet/test/peergroup/InetPeerGroupUtilsSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.scalatest.Matchers._
+
+import scala.concurrent.duration._
+class InetPeerGroupUtilsSpec extends FlatSpec with BeforeAndAfterAll {
+
+  implicit val patienceConfig: ScalaFutures.PatienceConfig = PatienceConfig(5 seconds)
+
+  behavior of "getChannelId"
+
+  it should "create a channelId for ip6address" in {
```
jtownson: Please use a table and remove copy/paste: http://www.scalatest.org/user_guide/table_driven_property_checks

shaileshp0110: Please review again resolved changed to Table driven

src/io/iohk/scalanet/test/peergroup/InetPeerGroupUtilsSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  it should "create a channelId for all ip6address and ip4Address" in {
+
+    forAll(values) { (ipAddress1, ipAddress2, port1, port2) =>
+      val remoteInet6Address = InetAddress.getByName(ipAddress1)
+      val localInet6Address = InetAddress.getByName(ipAddress2)
+      val remoteAddress = new InetSocketAddress(remoteInet6Address, port1)
+      val localAddress = new InetSocketAddress(localInet6Address, port2)
+
+      InetPeerGroupUtils.getChannelId(remoteAddress, localAddress).hashCode shouldBe a[Integer]
```
krcz: what does this test? can `.hashCode` method ever return non-`Integer`?

shaileshp0110: Hashcode always returns integer only, This was only to prove that getChannelD function won't throw an exception for Ipv6 addresses. It bit pointless test, I am not sure which is the other way to test

AlexITC: Just call `succeed` instead.

PR #31, opened 2019-06-13
A-667: Identify and fix server channel reception issue

EzequielPostan:
Problem:
- UDPPeergroup returns a monix PublishSubject to expose the stream of channels that a peer receives to communicate with clients.
  This has the characteristic that, no channel received before the subscription to the `server()` stream will actually be notified to the user.
- This commit changes the use of PublishSubject for ReplaySubject. The semantics of ReplaySubject is that the channels received before the
  subscription will arrive to the user. This is the same behaviour we have in the `in` method of UPD channels's implementation.

Consider
- This solves a requirement need, however, this could also create problems in other use cases.
  A proper solution should be discussed.

### DIFF
--- a/src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
+++ b/src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
@@ -45,4 +45,27 @@ class UDPPeerGroupSpec extends FlatSpec {
 
     isListeningUDP(pg1.config.bindAddress) shouldBe false
   }
+
+  it should "receive a channel and message that was sent before subscription to the server stream" in withTwoRandomUDPPeerGroups[
+    String
+  ] { (alice, bob) =>
+    val alicesMessage = Random.alphanumeric.take(1024 * 4).mkString
+
+    val aliceClient = alice.client(bob.processAddress).evaluated
+    aliceClient.sendMessage(alicesMessage).runAsync
+
+    expensiveComputation()
+
+    val bobFirstReceivedChannel = bob.server().headL.runAsync
+    bobFirstReceivedChannel.futureValue.to shouldBe alice.processAddress
+
+    val bobReceivedMessage = bobFirstReceivedChannel.futureValue.in.headL.runAsync
+    bobReceivedMessage.futureValue shouldBe alicesMessage
+  }
+
+  def expensiveComputation(): Unit = {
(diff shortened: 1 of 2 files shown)


marcesquerra (approved):

AlexITC (approved):
Looks good, it is worth considering whether we should receive the subject as an argument in order to let the users control what behavior they need, another way could be to just receive the strategy and create the stream with it.

src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
```diff
@@ -45,4 +45,27 @@ class UDPPeerGroupSpec extends FlatSpec {
...
+
+  it should "receive a channel and message that was sent before subscription to the server stream" in withTwoRandomUDPPeerGroups[
+    String
+  ] { (alice, bob) =>
+    val alicesMessage = Random.alphanumeric.take(1024 * 4).mkString
+
+    val aliceClient = alice.client(bob.processAddress).evaluated
+    aliceClient.sendMessage(alicesMessage).runAsync
+
+    expensiveComputation()
```
krcz, line 57: why not `Thread.sleep`, which wouldn't stress CPU needlessly?

krcz (approved):
It's ok as long as the number of connections is small. Otherwise we're keeping history of all connections, not only metadata, but whole `Channel` references, causing resource leak (but we have it in `activeChannels` anyway, so it doesn't matter here; if we start caring we'd have to refactor both things anyway).

src/io/iohk/scalanet/test/peergroup/UDPPeerGroupSpec.scala
EzequielPostan: I noticed that sometimes `Thread.sleep` produced strange behaviours and @loverdos mentioned that its use had also caused flaky tests in some CI tools. The intention was to be illustrative more than a final solution.

EzequielPostan:
@jtownson @shaileshp0110 
As you mentioned, this may need to be fixed in a different way. Both @AlexITC and @krcz mentioned the issues of receiving the whole history of messages before the subscription.
I will leave the PR open as part of the conversation, feel free to merge it or to define an specific issue to tackle this topic.

jtownson:
Superseded by https://iohk.myjetbrains.com/youtrack/issue/A-791.

PR #32, opened 2019-06-15
A 799/millify scalanet

jtownson:
Includes publishing, scoverage and decco dependency as a git submodule.


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "decco"]
```
AlexITC, line 1: Isn't there any other ways? I have always experienced a pain while dealing with submodules.

jtownson: Still looking for one :-(

.gitmodules
```diff
@@ -0,0 +1,5 @@
...
+	path = decco
+	url = git@github.com:input-output-hk/decco.git
```
loverdos: Why aren't we using the `https://github.com/input-output-hk/decco.git` URL?

.gitmodules
loverdos: Same comment as @AlexITC.

I can see this as an intermediate solution until the `decco` artifact is published somewhere, is that the idea?

.circleci/config.yml
```diff
@@ -7,6 +7,10 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: Initialize submodules
+          command: git submodule init && git submodule update --remote
```
loverdos: I believe you can use just one command:

```
git submodule update --init
```

Or

```
git submodule update --init --recursive
```

.circleci/config.yml
```diff
@@ -16,27 +20,16 @@ jobs:
...
-
-      - run:
-          name: build
-          command: |
-            bazel build --jobs=1 //src/io/iohk/scalanet:scalanet_EXT_deploy.jar
-
-            cp bazel-bin/src/io/iohk/scalanet/scalanet_EXT_deploy.jar scalanet.jar
-
-            .circleci/removeScalaStandardLibrary scalanet.jar
+          command: mill scalanet.test && mill scalanet.scoverage.htmlReport
```
loverdos, line 28: scoverage reports 89.69% coverage, looks good :-)

.gitmodules
jtownson: Yes, that's the idea.

.gitmodules
jtownson: I didn't realize this was the recommended approach. Changed to https.

.circleci/config.yml
jtownson: In fact, the submodule approach does not actually work (with either my or your commands). I've just replaced with a clone.

.circleci/config.yml
jtownson: Yes, Shailesh has written quite a few. We'll be playing the test pack story, so hopefully it will go upto 100% or very close.

jtownson:
Merged via command line.

PR #33, opened 2019-06-20
A 759/netty resource leek handling

shaileshp0110:


src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -116,10 +116,11 @@ object TCPPeerGroup {
     override val to: InetMultiAddress = InetMultiAddress(nettyChannel.remoteAddress())
 
     override def sendMessage(message: M): Task[Unit] = {
+      val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
       toTask({
         nettyChannel
-          .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
-      })
+          .writeAndFlush(byteBuf)
+      }).map(_ => byteBuf.release())
```
AlexITC: If the `writeAndFlush` method throws an exception, the buffer won't be released.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -302,10 +302,12 @@ object TLSPeerGroup {
     override val to: InetMultiAddress = InetMultiAddress(nettyChannel.remoteAddress())
 
     override def sendMessage(message: M): Task[Unit] = {
+      val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
+
       toTask(
         nettyChannel
           .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
-      )
+      ).map(_ => byteBuf.release())
```
AlexITC: The same.

src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -104,6 +105,7 @@ class UDPPeerGroup[M](val config: Config)(implicit codec: Codec[M], bufferInstan
                 activeChannels.put(channelId, channel)
                 channelSubject.onNext(channel)
                 messageE.foreach(message => channel.messageSubject.onNext(message))
+                datagram.content().release()
```
AlexITC, line 115: Shouldn't this be released after line 92 (after decoding the datagram content)?

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -302,10 +302,12 @@ object TLSPeerGroup {
     override val to: InetMultiAddress = InetMultiAddress(nettyChannel.remoteAddress())
 
     override def sendMessage(message: M): Task[Unit] = {
+      val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
+
       toTask(
         nettyChannel
           .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
```
AlexITC: You missed the `byteBuf`.

AlexITC (changes requested):
It looks like you would benefit from creating a function that creates the ByteBuffer and releases it after the computation completes, something like `withByteBuffer[A](...)(f: => Task[A]): Task[A]`

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
jtownson: This will leak when the task fails.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -216,6 +217,7 @@ object TCPPeerGroup {
       messageE.foreach { message =>
         messageSubject.onNext(message)
       }
+      byteBuf.release()
```
jtownson: This will leak if any exception gets thrown during decode. Think it should be
```
try {
 ...
} finally {
  byteBuf.release()
}
```

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -180,16 +180,16 @@ object TLSPeerGroup {
...
+      val messageE: Either[Codec.Failure, M] = codec.decode(byteBuf.nioBuffer().asReadOnlyBuffer())
       log.debug(
         s"Processing inbound message from remote address ${ctx.channel().remoteAddress()} " +
           s"to local address ${ctx.channel().localAddress()}, ${messageE.getOrElse("decode failed")}"
       )
-
       messageE.foreach { message =>
         messageSubject.onNext(message)
       }
+      byteBuf.release()
```
jtownson: As above.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
jtownson: as above

src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -68,6 +68,7 @@ class UDPPeerGroup[M](val config: Config)(implicit codec: Codec[M], bufferInstan
 
               val channel = activeChannels(channelId)
               messageE.foreach(message => channel.messageSubject.onNext(message))
+              datagram.content().release()
```
jtownson: should use try/finally for this too. onNext should not throw in theory but it could do in practice.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -116,10 +116,16 @@ object TCPPeerGroup {
...
-          .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
-      })
+      val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
+      try {
+        toTask({
+          nettyChannel
+            .writeAndFlush(byteBuf)
+        })
+      } finally {
+        byteBuf.release()
```
krcz: This is going to run before the task is finished, as the Task is going to be run in other thread. That means **byteBuf will be released before it is used**.

Please fix to:
```
val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
toTask {
  nettyChannel.writeAndFlush(byteBuf)
}.guarantee(Task(byteBuf.release()))
```

This way `byfeBuf` will be released when the task is ended, no matter how (completing successfully, failing, or being cancelled).

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -302,10 +307,15 @@ object TLSPeerGroup {
...
-          .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
-      )
+      val byteBuf = Unpooled.wrappedBuffer(codec.encode(message))
+      try {
+        toTask(
+          nettyChannel
+            .writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)))
+        )
+      } finally {
+        byteBuf.release()
```
krcz: `byfeBuf` is to be released before usage, please fix like before

krcz (changes requested):
Good catch on proper Netty buffers usage! Please fix the `Task` related issue and you'll have my approval.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
krcz: or even better, using special function for resources management (`Task.bracket`):
```
Task(Unpooled.wrappedBuffer(codec.encode(message))).bracket { byteBuf =>
  nettyChannel.writeAndFlush(byteBuf)
}(byteBuf => Task(byteBuf.release()))
```

jtownson: In fact, writeAndFlush releases the buffer. The writeAndFlush call returns a ChannelFuture and executes asynchronously. When wrapped in toTask so that it has completed, you receive an IllegalReferenceCountException from netty (because our release call tries to decrement the ref count to -1). I have updated these writeAndFlush spots accordingly.

src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
```diff
@@ -10,7 +10,7 @@ object InetPeerGroupUtils {
 
   type ChannelId = (InetSocketAddress, InetSocketAddress)
 
-  def toTask(f: util.concurrent.Future[_]): Task[Unit] = {
+  def toTask(f: => util.concurrent.Future[_]): Task[Unit] = {
     Task.create[Unit] { (_, cb) =>
       f.addListener(
         (future: util.concurrent.Future[_]) => if (future.isSuccess) cb.onSuccess(()) else cb.onError(future.cause())
```
AlexITC, line 16: I know this is not related to the PR but, why is this assuming that the future is already completed? if it wasn't successful doesn't mean that it has failed, it may have not completed yet.

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -206,15 +203,17 @@ object TCPPeerGroup {
...
-
-      val messageE: Either[Codec.Failure, M] = codec.decode(msg.asInstanceOf[ByteBuf].nioBuffer().asReadOnlyBuffer())
-      log.debug(
-        s"Processing inbound message from remote address ${ctx.channel().remoteAddress()} " +
-          s"to local address ${ctx.channel().localAddress()}, ${messageE.getOrElse("decode failed")}"
-      )
-
-      messageE.foreach { message =>
-        messageSubject.onNext(message)
+      val byteBuf = msg.asInstanceOf[ByteBuf]
```
AlexITC, line 206: Does `msg: ByteBuf` work? if it does, please remove the casting.

src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -180,15 +180,18 @@ object TLSPeerGroup {
...
-
-      val messageE: Either[Codec.Failure, M] = codec.decode(msg.asInstanceOf[ByteBuf].nioBuffer().asReadOnlyBuffer())
-      log.debug(
-        s"Processing inbound message from remote address ${ctx.channel().remoteAddress()} " +
-          s"to local address ${ctx.channel().localAddress()}, ${messageE.getOrElse("decode failed")}"
-      )
-
-      messageE.foreach { message =>
-        messageSubject.onNext(message)
+      val byteBuf = msg.asInstanceOf[ByteBuf]
```
AlexITC, line 183: The same.

AlexITC (approved):

src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
krcz: This is inside listener added by `.addListener` which is called only when the future is completed

src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
krcz: msg type is `Any`, so cast seems necessary

AlexITC: It could be worth to do pattern matching instead and throw a custom exception when the cast fails.

PR #34, opened 2019-06-21
POC for A-766

EzequielPostan:
This is a proof of concept to discuss what we would like to have as a test pack for peer groups.

### DIFF
--- a/src/io/iohk/scalanet/test/peergroup/PeerGroupTestPack.scala
+++ b/src/io/iohk/scalanet/test/peergroup/PeerGroupTestPack.scala
@@ -1,0 +1,107 @@
+package io.iohk.scalanet.peergroup
+
+import java.nio.ByteBuffer
+
+import io.iohk.decco.BufferInstantiator
+import io.iohk.scalanet.TaskValues._
+import monix.execution.Scheduler
+import org.scalatest.FlatSpec
+import org.scalatest.Matchers._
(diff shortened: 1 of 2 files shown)


jtownson (commented):
Think the responsibilities should be defined more strictly as
* Specs for concrete peer group classes create peer groups in various configurations.
* the test pack should define the test cases for in terms of the PeerGroup trait.
* The concrete peer group specs can then decide which tests to invoke from the test pack, passing the peer groups they have created.

Consider an example from the DTLSPeerGroupSpec.
```scala
object TestPack {
  // test pack defines a messaging test in terms of the PeerGroup trait
  def messagingTest[A](alice: PeerGroup[A, String], bob: PeerGroup[A, String]) = ???

  // standard tests for
  // initialization
  // shutdown
  // multicast
  // multiplexing
  // addressing
}

class DTLSPeerGroupSpec {
  // the concrete testcase 
  // declares that the DTLSPeerGroup supports messaging
  // creates DTLSPeerGroups with relevant configs (raw public keys and signed certificates)...
  // invokes the relevant test from the test pack
  it should "send and receive a message" in withTwoDTLSPeerGroups[String](rawKeyConfig, signedCertConfig) { (alice, bob) => 
    messagingTest(alice, bob)
  }

  def rawKeyConfig(...): DTLSPeerGroup.Config = // ...

  def signedCertConfig(...): DTLSPeerGroup.Config = // ...
}
```

PR #35, opened 2019-06-23
A-760/tcp reassembly

jtownson:
This is a PR showing how we might externalize the framing logic (currently handling by `LengthFieldPrepender`) from TCPPeerGroup and family.

The basic requirement is keep the encode/decode symmetry whilst removing the constraint that inbound buffers contain exactly one message.

I played around with Monix `Operator` and also a syntax sugar for decorating Observables. When I came to update the TCPPeerGroup, it turns out not to need those approaches but I have left them in with testcases for illustration.

There are a number of loose ends. Don't be shy! Point them out. Comments on the general approach also appreciated.


scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -0,0 +1,80 @@
...
+import io.iohk.decco.{BufferInstantiator, Codec}
+import io.iohk.decco.Codec.{DecodeResult, Failure}
+import io.iohk.scalanet.codec.FramingCodec.State._
+import io.iohk.scalanet.codec.FramingCodec._
+
+import scala.collection.mutable
+
+class FramingCodec[T](messageCodec: Codec[T]) extends StreamCodec[T] {
+
+  var state: State = LengthExpected
```
krcz: Could you make these private?

krcz (approved):
There is some mutability that could be avoided, but as this is low level network code, it is probably better that way, as it should be faster.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: Yes, of course.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -0,0 +1,79 @@
...
+object FramingCodec {
+
+  trait State
+
+  object State {
+    case object LengthExpected extends State
+    case object BytesExpected extends State
+  }
+
+  def read(n: Int, from: ByteBuffer, to: ByteBuffer): Unit = {
```
AlexITC: Don't you need to check that `from` has at least `n` bytes, and `to` has space for storing `n` bytes?

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  private var state: State = LengthExpected
+  private var length: Int = 0
+  private var db: ByteBuffer = _
+  private val nlb = ByteBuffer.allocate(4)
+
+  override def streamDecode[B](source: B)(implicit bi: BufferInstantiator[B]): Seq[T] = this.synchronized {
+    val b = bi.asByteBuffer(source)
+    val s = mutable.ListBuffer[T]()
+    while (b.remaining() > 0) {
```
AlexITC: I would use `hasRemaining` instead, it makes things clear, possibly, I can think that `remaining` method could take linear time while `hasRemaining` could take constant time but this is just an speculation.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -0,0 +1,79 @@
...
+      state match {
+        case LengthExpected =>
```
AlexITC, line 24: Wasn't this what the `LengthFieldPrepender` was doing and the task was supposed to avoid?

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -0,0 +1,79 @@
...
+import scala.collection.mutable
+
+class FramingCodec[T](messageCodec: Codec[T]) extends StreamCodec[T] {
+
+  private var state: State = LengthExpected
+  private var length: Int = 0
+  private var db: ByteBuffer = _
+  private val nlb = ByteBuffer.allocate(4)
+
+  override def streamDecode[B](source: B)(implicit bi: BufferInstantiator[B]): Seq[T] = this.synchronized {
```
AlexITC, line 19: If I understand correctly, in the case of corrupted bytes where you receive a length but there aren't enough bytes to consume, the outcome is that you will get an empty list.

If there are enough bytes to satisfy the expected length but those bytes are corrupted, you just don't get `T` which ends up being another empty list.

Wouldn't it be worth to log a trace on these scenarios?

AlexITC (approved):

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: Agreed. `hasRemaining` reads better. I had a look at the implementations and actually each is constant time:
```
    public final int remaining() {
        return limit - position;
    }
    public final boolean hasRemaining() {
        return position < limit;
    }
```

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: The required checks are in the streamDecode method. What I've done is rename to `readUnchecked` and mark it as private. Hopefully then it's clear that it's just part of the implementation.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: The idea is that the delimiting is not hard-coded in TCPPeerGroup so that now the user has complete control over the wire format.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: All the bytes coming from the TCP stack will have been checksummed so the corruption should not happen.

What could happen is the connection could break somewhere during a read but that should be handled by the code managing the connection.

What I have noticed is that it would be very easy to send a rogue length parameter then try to fill the nodes memory with random bytes. I've added a maxBufferSize parameter to stop this (this was something supported by LengthFieldDecoder).

jtownson:
Merged via command line.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
AlexITC: I was not referring to corruption while sending bytes on the wire but corruption on purpose, thanks.

PR #36, opened 2019-06-24
A-793 Publish scalanet into Sonatype

krcz:


.circleci/publish
```diff
@@ -0,0 +1,12 @@
...
+
+set -euv
+
+echo $GPG_KEY | base64 --decode | gpg --batch --import
+
+gpg --passphrase $GPG_PASSPHRASE --batch --yes -a -b LICENSE
+
+mill mill.scalalib.PublishModule/publishAll \
+    "$OSS_USERNAME":"$OSS_PASSWORD" \
+    "$GPG_PASSPHRASE" \
```
AlexITC, line 11: As this is a public repository, what prevents people from submitting a PR containing code like `println(java.lang.System.getenv("GPG_PASSPHRASE"))` and looking into the logs to see the printed key?

krcz: That's a very good question! Using environment variables is [one of two suggested solutions for storing secrets](https://support.circleci.com/hc/en-us/articles/360006717953-Storing-secret-files-certs-etc-). Forked pull request builds are currently disabled for scalanet and even when we enable them it is [possible to disable passing secrets to them](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests) and that is the default option.

But if you prefer I can change the setup to use contexts, which allow restricting environment variables usage to particular users.

jtownson (changes requested):
Unfortunately, this change breaks scoverage (I think, it is because of the way the `moduleDeps` is being overridden from both PublishModule and ScoverageModule).

krcz:
> Unfortunately, this change breaks scoverage (I think, it is because of the way the `moduleDeps` is being overridden from both PublishModule and ScoverageModule).

That is a bug in mill: https://github.com/lihaoyi/mill/issues/620

I thought we can disable SCoverage until it is fixed, but if you don't want such solution I can try to fix mill.

.circleci/publish
AlexITC: > Forked pull request builds are currently disabled

That looks like the right choice from my side, thanks for the details.

krcz: To be honest I haven't thought about that issue, I just used proposed solution. Only good CircleCI saved me from introducing a security bug.

build.sc
```diff
@@ -2,16 +2,27 @@ import mill._
...
-
-object scalanet extends ScoverageModule {
+// ScoverageModule creates bug when using custom repositories:
+// https://github.com/lihaoyi/mill/issues/620
+// object scalanet extends ScalaModule with PublishModule {
+object scalanet extends ScalaModule with PublishModule with ScoverageModule {
 
   def scalaVersion = "2.12.7"
 
+  def publishVersion = "1.0-SNAPSHOT"
```
jtownson: Please make this 0.1-SNAPSHOT

jtownson (approved):
Okay, fine to merge without scoverage. Please see version number change. (note Shailesh's A-796-Use-Decco-As-Sontatype-Dependency branch may cause conflicts)

PR #37, opened 2019-06-26
Add InMemoryPeerGroup

EzequielPostan:
The idea for this peer group is to aid testing/design for both the users and scalanet developers.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
@@ -9,37 +9,12 @@ import InMemoryPeerGroup._
 import scala.collection.concurrent.TrieMap
 import scala.util.Random
 
-// For simplicity we can model the network with three main interactions.
-// 1. Register to the network.
-// 2. Disconnect from the network.
-// 3. Send messages.
-class Network[A, M] {
-  private val peers: TrieMap[A, InMemoryPeerGroup[A, M]] = TrieMap()
-
-  def clear(): Unit = peers.clear()
-
-  def register(peer: InMemoryPeerGroup[A, M]): Result[Unit] =
-    if (peers.get(peer.processAddress).isEmpty) {
-      peers += (peer.processAddress -> peer)
-      allGood(())
-    } else error("The address is already in use")
-
-  def disconnect(peer: InMemoryPeerGroup[A, M]): Unit = peers -= peer.processAddress
-
-  def deliverMessage(channelID: ChannelID, from: A, to: A, message: M): Result[Unit] =
-    peers.get(to) match {
-      case None => error(s"Unreachable peer $to")
-      case Some(destination) =>
-        allGood(destination.receiveMessage(channelID, from, message))
-    }
-}
-
 class InMemoryPeerGroup[A, M](address: A)(implicit network: Network[A, M]) extends PeerGroup[A, M] {
 
   private var status: PeerStatus = PeerStatus.NotInitialized
   private val channelStream = ConcurrentSubject[InMemoryChannel[A, M]](MulticastStrategy.publish)
   private val channelsMap: TrieMap[ChannelID, InMemoryChannel[A, M]] = TrieMap()
-  def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
+  private def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
     channelsMap.get(channelID) match {
       case None =>
         val newChannel = new InMemoryChannel(channelID, processAddress, from)
@@ -52,27 +27,27 @@ class InMemoryPeerGroup[A, M](address: A)(implicit network: Network[A, M]) exten
   }
 
   // Public interface
-  def processAddress: A = address
-  def initialize(): Result[Unit] =
+  override def processAddress: A = address
+  override def initialize(): Result[Unit] =
     if (status == PeerStatus.Listening) error("Peer already connected")
     else {
       status = PeerStatus.Listening
       network.register(this)
     }
-  def client(to: A): Result[InMemoryChannel[A, M]] = {
+  override def client(to: A): Result[InMemoryChannel[A, M]] = {
     val channelID = newChannelID()
     val newChannel = new InMemoryChannel[A, M](channelID, processAddress, to)
     channelsMap += (channelID -> newChannel)
     allGood(newChannel)
   }
-  def server(): Observable[InMemoryChannel[A, M]] = {
+  override def server(): Observable[InMemoryChannel[A, M]] = {
     status match {
       case PeerStatus.NotInitialized => throw new Exception(s"Peer $processAddress is not initialized yet")
       case PeerStatus.Listening => channelStream
       case PeerStatus.ShutDowned => throw new Exception(s"Peer $processAddress was shut downed")
     }
   }
-  def shutdown(): Result[Unit] = {
+  override def shutdown(): Result[Unit] = {
     status match {
       case PeerStatus.NotInitialized =>
         allGood(()) // Should we return an error here?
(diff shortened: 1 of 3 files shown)


scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+import InMemoryPeerGroup._
+
+import scala.collection.concurrent.TrieMap
+import scala.util.Random
+
+// For simplicity we can model the network with three main interactions.
+// 1. Register to the network.
+// 2. Disconnect from the network.
+// 3. Send messages.
+class Network[A, M] {
```
AlexITC: Create a separate file unless there is good reason not to.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+  private val peers: TrieMap[A, InMemoryPeerGroup[A, M]] = TrieMap()
+
+  def clear(): Unit = peers.clear()
+
+  def register(peer: InMemoryPeerGroup[A, M]): Result[Unit] =
+    if (peers.get(peer.processAddress).isEmpty) {
```
AlexITC: While it is unlikely to happen, is there a simple way to do a compare and set? if there is, it is worth using it.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+        allGood(destination.receiveMessage(channelID, from, message))
+    }
+}
+
+class InMemoryPeerGroup[A, M](address: A)(implicit network: Network[A, M]) extends PeerGroup[A, M] {
+
+  private var status: PeerStatus = PeerStatus.NotInitialized
+  private val channelStream = ConcurrentSubject[InMemoryChannel[A, M]](MulticastStrategy.publish)
+  private val channelsMap: TrieMap[ChannelID, InMemoryChannel[A, M]] = TrieMap()
+  def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
```
AlexITC: If this is not supposed to go in the public interface, then make it private.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+        channelsMap += (channelID -> newChannel)
+        newChannel.depositMessage(msg)
+        channelStream.onNext(newChannel)
+      case Some(ch) =>
+        ch.depositMessage(msg)
+    }
+  }
+
+  // Public interface
+  def processAddress: A = address
```
AlexITC: I bet this needs the `override`

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+  def shutdown(): Result[Unit] = {
+    status match {
+      case PeerStatus.NotInitialized =>
+        allGood(()) // Should we return an error here?
+      case PeerStatus.Listening =>
+        status = PeerStatus.ShutDowned
+        network.disconnect(this)
+        allGood(())
+      case PeerStatus.ShutDowned =>
+        error("The peer group was already down") // Should we return an error here?
```
AlexITC: Isn't this actually returning an error?

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+  sealed trait PeerStatus
+  object PeerStatus {
+    case object NotInitialized extends PeerStatus
+    case object Listening extends PeerStatus
+    case object ShutDowned extends PeerStatus
+  }
+
+  trait ChannelStatus
+  object ChannelStatus {
+    case object Open extends ChannelStatus
```
AlexITC: `Opened` and `Closed` sounds adequate to me.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -0,0 +1,142 @@
...
+    case object Close extends ChannelStatus
+  }
+
+  type Result[A] = Task[A]
+  def allGood[A](x: A): Result[A] = Task.now(x)
+  def error(msg: String): Result[Nothing] = Task.now(throw new Exception(msg))
```
AlexITC: Is this the correct way to initialize an always-failing task? I'm not very familiar with monix but I would expect something like `Future.failed`

scalanet/test/src/io/iohk/scalanet/peergroup/InMemoryPeerGroupSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+import io.iohk.scalanet.TaskValues._
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{BeforeAndAfter, FlatSpec}
+import org.scalatest.Matchers._
+import org.scalatest.concurrent.ScalaFutures._
+
+import scala.concurrent.Future
+import scala.util.Random
+
+class InMemoryPeerGroupSpec extends FlatSpec with BeforeAndAfter {
```
AlexITC, line 12: It looks to me like we could merge most *PeerGroupSpec and test all existing implementations.

AlexITC (approved):

scalanet/test/src/io/iohk/scalanet/peergroup/InMemoryPeerGroupSpec.scala
EzequielPostan: that is the idea of another ticket and also why I played with the creation of the PeerUtils typeclass

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: yes, the question in the comment was part of an initial attempt to unify the behaviour of PeerGroups. This PeerGroup started as an attempt of an executable specification (and is intentionally mainly that). On the process of coding it, I realised it could actually be a useful PeerGroup.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: abstract methods can be overriden without the need of the override modifier. However, I do tend to always add the modifier. Good catch.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: I will check if I can make it only accessible to the Network class.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: The thing is that I am not just interested in comparing and setting. I want to actually raise an error if the comparison fails.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: This is internal/relevant only to this PeerGroup. I could attempt to make the Network part of the companion object.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: it seems that `Task.raiseError` can do the trick. I will try it.

EzequielPostan:
Thank you for the review @AlexITC

PR #38, opened 2019-06-27
A-763/errors

jtownson:
This PR handles
* connection and sending errors in TCP and TLS
* message size errors in UDP and DTLS
* client handshake errors
* failed connection events reports to servers.
(There is also a little bit of test refactoring in preparation for the test pack story).


.gitmodules~
```diff
@@ -0,0 +1,5 @@
+[submodule "decco"]
```
AlexITC: Isn't this already in sonatype?

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -5,22 +5,23 @@ import monix.reactive.{MulticastStrategy, Observable}
...
 import scala.collection.concurrent.TrieMap
-import scala.util.Random
 
 class InMemoryPeerGroup[A, M](address: A)(implicit network: Network[A, M]) extends PeerGroup[A, M] {
 
-  private var status: PeerStatus = PeerStatus.NotInitialized
-  private val channelStream = ConcurrentSubject[InMemoryChannel[A, M]](MulticastStrategy.publish)
-  private val channelsMap: TrieMap[ChannelID, InMemoryChannel[A, M]] = TrieMap()
-  private def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
+  var status: PeerStatus = PeerStatus.NotInitialized
```
AlexITC: This is dangerous, why aren't these private anymore? can't you use package private instead?

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -121,8 +124,67 @@ object PeerGroup {
...
+      def collector[A]: PartialFunction[ServerEvent[A, _], HandshakeException[A]] = {
+        case HandshakeFailed(failure) => failure
+      }
+    }
+
+    implicit class ServerOps[A, M](observable: Observable[ServerEvent[A, M]]) {
+      def collectChannelCreated: Observable[Channel[A, M]] = observable.collect(ChannelCreated.collector)
+      def collectHandshakeFailure: Observable[HandshakeException[A]] = observable.collect(HandshakeFailed.collector)
+    }
+//    def skip[R, T](r: R, t: T): R = r
```
AlexITC: Remove useless code.

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
```diff
@@ -45,12 +46,12 @@ class SimplePeerGroup[A, AA, M](
...
-    underLyingPeerGroup.server().map { underlyingChannel: Channel[AA, Either[ControlMessage[A, AA], M]] =>
-      val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
-      val a = reverseLookup(underlyingChannel.to)
-      debug(s"Received new server channel from $a")
-      new ChannelImpl(a, List(underlyingChannel))
+  override def server(): Observable[ServerEvent[A, M]] = {
+    underLyingPeerGroup.server().collectChannelCreated.map {
+      underlyingChannel: Channel[AA, Either[ControlMessage[A, AA], M]] =>
+        val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
+        val a = reverseLookup(underlyingChannel.to)
```
AlexITC, line 55: can't reverse lookup fail?

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -1,12 +1,14 @@
...
+import java.io.IOException
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
 import java.util.concurrent.ConcurrentHashMap
 
 import io.iohk.decco.{BufferInstantiator, Codec}
-import io.iohk.scalanet.peergroup.PeerGroup.TerminalPeerGroup
-import io.iohk.scalanet.peergroup.InetPeerGroupUtils.{getChannelId, toTask}
+import io.iohk.scalanet.peergroup.PeerGroup.{ChannelSetupException, MessageMTUException, ServerEvent, TerminalPeerGroup}
+import InetPeerGroupUtils.{ChannelId, getChannelId, toTask}
```
AlexITC: Use fully qualified import instead.

AlexITC (approved):

.gitmodules~
jtownson: Arrrrgh, it's the backup file created by emacs or vi. Shouldn't be in git. Removed.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
jtownson: I've made them private to the package.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
jtownson: Thanks

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
jtownson: Yep, it can. Actually, we'll be removing this when we take out InetMultiAddress.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
jtownson: thanks

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
```diff
@@ -45,12 +46,12 @@ class SimplePeerGroup[A, AA, M](
...
   }
 
-  override def server(): Observable[Channel[A, M]] = {
-    underLyingPeerGroup.server().map { underlyingChannel: Channel[AA, Either[ControlMessage[A, AA], M]] =>
-      val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
-      val a = reverseLookup(underlyingChannel.to)
-      debug(s"Received new server channel from $a")
-      new ChannelImpl(a, List(underlyingChannel))
+  override def server(): Observable[ServerEvent[A, M]] = {
+    underLyingPeerGroup.server().collectChannelCreated.map {
```
krcz: Are you using just channel creations, silently ignoring other server events? Is it possible to expose handshake failures somehow?

scalanet/src/io/iohk/scalanet/peergroup/StaticAddressMappedPeerGroup.scala
```diff
@@ -27,10 +29,14 @@ class StaticAddressMappedPeerGroup[A, AA, M](
       new ChannelImpl(to, underlyingChannel)
     }
 
-  override def server(): Observable[Channel[A, M]] = {
+  override def server(): Observable[ServerEvent[A, M]] = {
     underLyingPeerGroup
       .server()
-      .map(underlyingChannel => new ChannelImpl(reverseLookup(underlyingChannel.to), underlyingChannel))
+      .collectChannelCreated
```
krcz: As before: is it possible to map other types of events as well?

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -116,6 +123,10 @@ object TCPPeerGroup {
 
     override def sendMessage(message: M): Task[Unit] = {
       toTask(nettyChannel.writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)(bi))))
+        .onErrorRecoverWith {
+          case e: IOException =>
+            Task(throw new ChannelBrokenException[InetMultiAddress](to, e))
```
krcz: I think it's better to use `Task.raiseError(new ...)`, as it is pure data operation - it doesn't involve context switching and invoking real JVM throw-catch mechanism.

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -166,10 +177,16 @@ object TCPPeerGroup {
         }
       })
 
-    def initialize: Task[ClientChannelImpl[M]] = toTask(bootstrap.connect(inetSocketAddress)).map(_ => this)
+    def initialize: Task[ClientChannelImpl[M]] = {
+      toTask(bootstrap.connect(inetSocketAddress))
+        .onErrorRecoverWith {
+          case e: ConnectException =>
+            Task(throw new ChannelSetupException[InetMultiAddress](to, e))
```
krcz: Please use `Task.raiseError`

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
```diff
@@ -179,6 +196,10 @@ object TCPPeerGroup {
           )
           toTask(ctx.writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)(bi))))
         })
+        .onErrorRecoverWith {
+          case e: IOException =>
+            Task(throw new ChannelBrokenException[InetMultiAddress](to, e))
```
krcz: Please use `Task.raiseError`

krcz (changes requested):
I have few small comments. As for the `ServerEvent` mapping in wrapping peer groups it can be postponed into future, but as for `Task(throw ...)` vs `Task.raiseError(...)` please make changes before merging.

I haven't read into all changes yet, I'll return to this PR on Monday.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
```diff
@@ -5,22 +5,24 @@ import monix.reactive.{MulticastStrategy, Observable}
...
 
-  private var status: PeerStatus = PeerStatus.NotInitialized
-  private val channelStream = ConcurrentSubject[InMemoryChannel[A, M]](MulticastStrategy.publish)
-  private val channelsMap: TrieMap[ChannelID, InMemoryChannel[A, M]] = TrieMap()
-  private def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
+  private[peergroup] var status: PeerStatus = PeerStatus.NotInitialized
+  private[peergroup] val channelStream = ConcurrentSubject[ServerEvent[A, M]](MulticastStrategy.publish)
+  private[peergroup] val channelsMap: TrieMap[ChannelID, InMemoryChannel[A, M]] = TrieMap()
+
+  def receiveMessage(channelID: ChannelID, from: A, msg: M): Unit = {
```
EzequielPostan, line 19: I am not following, why couldn't the above members (`status`, `channelStream` and `channelMap`) be left as private and this particular method neither? All of those should be used only by the internal implementation and seems to not be used by any other entity in this PR.

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
EzequielPostan: is this just ignoring the error messages? shouldn't you handle them?

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
EzequielPostan: I think you could call Task.raiseError and avoid throwing the exception

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
EzequielPostan: same as before, Task.raiseError

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
EzequielPostan: same as before, Task.raiseError

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -240,23 +234,43 @@ object TLSPeerGroup {
...
+    }
+
     def initialize: Task[ClientChannelImpl[M]] = {
-      toTask(bootstrap.connect(inetSocketAddress)).map(_ => this)
+      toTask(bootstrap.connect(inetSocketAddress))
+        .flatMap(_ => Task.fromFuture(activationF))
+        .map(_ => this)
+        .onErrorRecoverWith {
+          case t: Throwable =>
+            Task.raiseError(mapException(t))
```
EzequielPostan, line 269: could this be re-written? 
In a way we have two futures that are converted into Tasks, then we are returning a Task with a reference to the object the client already used to call `initialise`, why couldn't we just return, something like Task.unit? avoiding the strange `map(_ => this)`
A related comment, isn't `activationF` evaluation running from the call to `activationF.future` (Line 210), what are we winning by adding it inside the Task that is discarding it anyway? Maybe I am misunderstanding Task semantics.

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -270,6 +284,10 @@ object TLSPeerGroup {
           )
           toTask(ctx.writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)(bi))))
         })
+        .onErrorRecoverWith {
+          case e: IOException =>
+            Task(throw new ChannelBrokenException[InetMultiAddress](to, e))
```
EzequielPostan: Task.raiseError

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -301,12 +320,32 @@ object TLSPeerGroup {
...
+      })
       .addLast(new MessageNotifier(messageSubject, codec, bi))
 
     override val to: InetMultiAddress = InetMultiAddress(nettyChannel.remoteAddress())
 
     override def sendMessage(message: M): Task[Unit] = {
       toTask(nettyChannel.writeAndFlush(Unpooled.wrappedBuffer(codec.encode(message)(bi))))
+        .onErrorRecoverWith {
+          case e: ClosedChannelException =>
+            Task(throw new ChannelBrokenException[InetMultiAddress](to, e))
```
EzequielPostan: Task.raiseError

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -149,9 +151,12 @@ class UDPPeerGroup[M](val config: Config)(implicit codec: Codec[M], bufferInstan
...
         nettyChannel: NioDatagramChannel
     ): Task[Unit] = {
-      toTask(
-        nettyChannel.writeAndFlush(new DatagramPacket(Unpooled.wrappedBuffer(codec.encode(message)), recipient, sender))
-      )
+      val encodedMessage = codec.encode(message)
+      toTask(nettyChannel.writeAndFlush(new DatagramPacket(Unpooled.wrappedBuffer(encodedMessage), recipient, sender)))
+        .onErrorRecoverWith {
+          case _: IOException =>
+            Task(throw new MessageMTUException[InetMultiAddress](to, encodedMessage.capacity()))
```
EzequielPostan: Task.raiseError

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -166,22 +171,26 @@ class UDPPeerGroup[M](val config: Config)(implicit codec: Codec[M], bufferInstan
...
-      activeChannels.put(channelId, channel)
-      channel
-    }
+        val channel = new ChannelImpl(nettyChannel, localAddress, to.inetSocketAddress, ReplaySubject[M]())
+        activeChannels.put(channelId, channel)
+        channel
+      }
+      .onErrorRecoverWith {
+        case e: Throwable =>
+          Task(throw new ChannelSetupException[InetMultiAddress](to, e))
```
EzequielPostan: Task.raiseError

scalanet/test/src/io/iohk/scalanet/peergroup/InMemoryPeerGroupSpec.scala
```diff
@@ -53,15 +55,15 @@ class InMemoryPeerGroupSpec extends FlatSpec with BeforeAndAfter {
       bobServer.futureValue.to shouldBe alice.processAddress
     }
 
-  it should "not receive a channel that started before the subscription to the server() stream" in
+  ignore should "not receive a channel that started before the subscription to the server() stream" in
```
EzequielPostan, line 58: is this the test we noticed to be flaky?

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
```diff
@@ -45,12 +46,12 @@ class SimplePeerGroup[A, AA, M](
...
+      underlyingChannel: Channel[AA, Either[ControlMessage[A, AA], M]] =>
+        val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
```
krcz: Not related to this ticket, note for the future: just iterating over elements here, searching for value would be faster here (O(n) vs O(n log n) for tree based maps); maintaining updated reverse map would be even faster, but prone to race conditions

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -240,23 +234,43 @@ object TLSPeerGroup {
...
-                super.userEventTriggered(ctx, evt)
               }
             })
             .addLast(new MessageNotifier[M](messageSubject, codec, bi))
         }
       })
 
+    private def mapException(t: Throwable): Throwable = t match {
+      case _: ClosedChannelException =>
+        new peergroup.PeerGroup.ChannelBrokenException[InetMultiAddress](to, t)
```
krcz: Is it possible to use just `PeerGroup.ChannelBrokenException` or even `ChannelBrokenException`?

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -240,23 +234,43 @@ object TLSPeerGroup {
...
+      case _: SSLKeyException =>
+        new iohk.scalanet.peergroup.PeerGroup.HandshakeException[InetMultiAddress](to, t)
+      case _ =>
+        t
+    }
+
     def initialize: Task[ClientChannelImpl[M]] = {
-      toTask(bootstrap.connect(inetSocketAddress)).map(_ => this)
+      toTask(bootstrap.connect(inetSocketAddress))
+        .flatMap(_ => Task.fromFuture(activationF))
```
krcz, line 265: where is `activationF` related promise fulfilled? it's hard to tell where does it come from - do we need to rely on state here?

krcz, line 265: where is `activationF` related promise fulfilled? it's hard to tell where does it come from - do we need to rely on state here?

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
```diff
@@ -301,12 +320,32 @@ object TLSPeerGroup {
...
+        override def userEventTriggered(ctx: ChannelHandlerContext, evt: Any): Unit = {
+          evt match {
+            case e: SslHandshakeCompletionEvent =>
+              val localAddress = InetMultiAddress(ctx.channel().localAddress().asInstanceOf[InetSocketAddress])
+              val remoteAddress = InetMultiAddress(ctx.channel().remoteAddress().asInstanceOf[InetSocketAddress])
+              log.debug(
+                s"Ssl Handshake server channel from $localAddress " +
+                  s"to $remoteAddress with channel id ${ctx.channel().id} and ssl status ${e.isSuccess}"
+              )
+              if (!e.isSuccess)
```
krcz, line 330: please use brackets when if branch statement has side effects (here it does)

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
jtownson: True. I've also mapped the handshake failures.

scalanet/src/io/iohk/scalanet/peergroup/StaticAddressMappedPeerGroup.scala
jtownson: Ditto.

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
jtownson: Agreed.

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
jtownson: Agreed. I have mapped the handshake failure too.

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
jtownson: I moved the creation of test instances into the test directory which then requires access to the internal state. We can discuss PeerUtils et al a bit later.

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
jtownson: Agreed.

scalanet/src/io/iohk/scalanet/peergroup/TCPPeerGroup.scala
jtownson: Yes.

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
jtownson: Hmm, it is a bit of a mess, I agree. Had a play around and could not see anything much cleaner because there are 3 separate async constructs (netty's Future, scala's Future and Task). Think I will raise a story to upgrade monix (the next minor version picks up a bunch of new ways to create tasks) and then get rid of the scala Promises. Sound reasonable?

scalanet/test/src/io/iohk/scalanet/peergroup/InMemoryPeerGroupSpec.scala
jtownson: Yes.

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
jtownson: Noted. Thanks.

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
```diff
@@ -47,11 +47,15 @@ class SimplePeerGroup[A, AA, M](
   }
 
   override def server(): Observable[ServerEvent[A, M]] = {
-    underLyingPeerGroup.server().collectChannelCreated.map {
-      underlyingChannel: Channel[AA, Either[ControlMessage[A, AA], M]] =>
-        val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
+    // FIXME pt1 addressing done as part of own proto
+    // FIXME pt2 somehow addressing as QoS option in underlying peer group allows fallback to underlying addressing?
+    val reverseLookup: mutable.Map[AA, A] = routingTable.map(_.swap)
```
krcz, line 52: Can one underlying address (`AA` one) be assigned to many multicast ones (`A`)? What do we want to do in such case? Because now it seems that we are choosing random multicast address such that underlying incoming address is assigned to.

~But I feel like that might be just one of issues possibly related to multicasting, I believe that we should create theoretic model for that and define contracts - we should create a separate task for that.~ [EDIT] that was my misunderstanding, sorry

scalanet/src/io/iohk/scalanet/peergroup/StaticAddressMappedPeerGroup.scala
```diff
@@ -30,13 +30,13 @@ class StaticAddressMappedPeerGroup[A, AA, M](
...
-      .server()
-      .collectChannelCreated
-      .map { underlyingChannel =>
+    underLyingPeerGroup.server().map {
+      case ChannelCreated(underlyingChannel) =>
         val a = reverseLookup(underlyingChannel.to)
         ChannelCreated(new ChannelImpl(a, underlyingChannel))
-      }
+      case HandshakeFailed(failure) =>
+        HandshakeFailed[A, M](new HandshakeException[A](reverseLookup(failure.to), failure.cause))
```
krcz, line 38: I guess `reverseLookup` can fail here as well - in case of unknown node connecting - should we throw exception here? Maybe silently (or with logging) ignore the connection?

krcz (approved):

scalanet/src/io/iohk/scalanet/peergroup/SimplePeerGroup.scala
krcz: During Slack conversation with Jeremy I understood that `routingTable` stores just mapping for single node addresses, not multicast ones, so there shouldn't be two keys with the same value (i.e. two `AA` underlying addresses with the same `A` one in `routingTable`, but they might exist in `multiCastTable`).

scalanet/src/io/iohk/scalanet/peergroup/TLSPeerGroup.scala
EzequielPostan: It partly addresses the comment, two minor pending question are:
- Why return a reference to `this` inside the returned Task value? Couldn't you return Task[Unit]?
- Also, why involving the Task that calls the promise's future? My remark there was that it seems to be adding no purpose (as the future was already running and wrapping it inside the Task seems to add no effect).

scalanet/src/io/iohk/scalanet/peergroup/InMemoryPeerGroup.scala
EzequielPostan: I see, I am fine to leave the proper solution for the test pack PR, but I will require for these methods to be private again in the final solution.

jtownson:
Merged via command line.

PR #39, opened 2019-07-02
A-760/tcp reassembly

jtownson:
This has a bug fix from the previous PR. Overlength frames are now ignored but consumed without buffering so that messaging decoding multiple frames remains aligned.


scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
```diff
@@ -84,4 +109,12 @@ object FramingCodec {
       m += 1
     }
   }
+
+  private def skipUnchecked(n: Int, from: ByteBuffer): Unit = {
+    var m: Int = 0
```
krcz: you could use `.position(int)` method to move position in one operation, instead of incrementally

scalanet/test/src/io/iohk/scalanet/NetUtils.scala
```diff
@@ -64,6 +64,13 @@ object NetUtils {
     a
   }
 
+  def concat(bs: ByteBuffer*): ByteBuffer = {
```
krcz: Is it used anywhere?

krcz (approved):
I think adding a flag that allowed to fail on oversized frames instead of skipping them silently might be a good idea, but that's just an idea for the future.

scalanet/src/io/iohk/scalanet/codec/FramingCodec.scala
jtownson: Agreed. Changed.

scalanet/test/src/io/iohk/scalanet/NetUtils.scala
jtownson: No. Deleted.

jtownson:
> I think adding a flag that allowed to fail on oversized frames instead of skipping them silently might be a good idea, but that's just an idea for the future.

`maxFrameSize = Int.maxValue` (the default value) achieves the same effect.

jtownson:
Merged via command line.

PR #43, opened 2019-07-24
A 724/kademlia discovery

jtownson:
This is the first basic implementation of a PeerGroup backed by a kademlia discovery process.


scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+  def contains(nodeId: BitVector): Boolean = {
+    buckets.getOrElse(Xor.d(nodeId, this.baseId), Set.empty[BitVector]).contains(nodeId)
+  }
+
+  /**
+    * Iterate over all elements in the KBuckets.
+    *
+    * @return an iterator
+    */
+  def iterator: Iterator[BitVector] = new Iterator[BitVector] {
```
krcz: Why such wrapping instead of just `= buckets.flatMap(_._2).iterator`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+    * @return
+    */
+  def remove(nodeId: BitVector): KBuckets = {
+    buckets
+      .get(Xor.d(nodeId, this.baseId))
+      .foreach(
+        bucket =>
+          bucket
+            .find(_ == nodeId)
+            .foreach(record => bucket.remove(record))
```
krcz: Does `remove` require presence of the element? If not why are you using `find` and then `remove` instead of just `remove`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+    override def next(): BitVector = flatIterator.next()
+  }
+
+  /**
+    * Remove an element by id.
+    *
+    * @param nodeId the nodeId to remove
+    * @return
+    */
+  def remove(nodeId: BitVector): KBuckets = {
```
krcz: In my opinion it's better not to return value from `remove` when mutating the collection as that might give a feeling that the original collection is not mutated.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+
+    override def findNodes(to: A, message: FindNodes[A]): Future[Nodes[A]] = {
+      peerGroup
+        .client(to)
+        .flatMap { channel =>
+          channel.sendMessage(message).runAsync.onComplete {
+            case Failure(exception) =>
+              log.info(s"findNodes request to $to failed to send. Got error: $exception")
```
krcz: If sending fails, should we still wait for response? few lines down?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -0,0 +1,52 @@
...
+        channel.in.collect {
+          case f @ FindNodes(_, _, _, _) =>
+            (channel, f)
+        }
+      }
+    }
+
+    override def findNodes(to: A, message: FindNodes[A]): Future[Nodes[A]] = {
+      peerGroup
+        .client(to)
```
krcz: Does this create a new channel for each `findNodes` call?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -0,0 +1,192 @@
...
+  private val log = LoggerFactory.getLogger(getClass)
+
+  add(config.nodeId, config.nodeRecord)
+
+  config.knownPeers.foreach {
+    case (peerId, peerRecord) => add(peerId, peerRecord)
+  }
+
+  info("Executing initial enrolment process to join the network...")
+  Try(Await.result(lookup(config.nodeId), 2 seconds)).toEither match {
```
krcz: Is `Await` really necessary? Couldn't we have companion object method for constructing `KRouter`, returning `Task[KRouter[V]]`, running the necessary enrolment and then injecting the results into real `KRouter` constructor?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -0,0 +1,192 @@
...
+            kNodes.nodes.foreach { case (id, record) => add(id, record) }
+
+            val kIds = kNodes.nodes.map(_._1)
+
+            val closestNext: Seq[BitVector] =
+              (closestNodes ++ kIds).sorted(xorOrdering)
+
+            val querySetNext: Set[BitVector] = querySet + knownNode
+
+            loop(querySetNext, closestNext)
```
krcz: Are you confident that such procedure will query each node at most once? It looks that computation branches, launching a `loop` subcomputation for each `knowNode`. Each subcomputation has their own `querySet` (which as I understand is set of already queried nodes), this value is not shared. Nothing stops different branches from querying the same node.

krcz (changes requested):
I've left some some comments in the code, but I also have more general one: does constructing `KPeerGroup` require two underlying peer groups: one for the router and the other for communication itself?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/Xor.scala
```diff
@@ -0,0 +1,23 @@
+package io.iohk.scalanet.peergroup.kademlia
+
+import scodec.bits.BitVector
+
+object Xor {
+
+  def d(a: BitVector, b: BitVector): Int = {
+    assert(a.length == b.length)
+    a.length.toInt - leadingZeros(a.xor(b))
```
EzequielPostan: as discussed before, this is not Kademlia's paper metric. It does not satisfy unidirectionality.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+import scodec.bits.BitVector
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+class KBuckets(val baseId: BitVector) {
+
+  private val buckets = new ConcurrentHashMap[Int, KBucket]().asScala
+
+  add(baseId)
```
EzequielPostan, line 18: minor comment, in the paper, the node does not store its own id in its KBuckets.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+    // each kbucket covers a distance range [2^i, 2^i+1 - 1]
+    // this means that buckets with larger i cover larger ranges.
+    // so, for eg, bucket 1 covers 4-2-1=1 node whereas bucket 7 covers 256-128-1=127 nodes
+    // in general each bucket covers 2^i+1 - 2^i - 1 = 2^i - 1
+
+    // therefore when finding an arbitrary nodeId
+    // find the prefix
+
+    // replace with routing tree described in kademlia paper...
+    iterator.toList.sorted(new XorOrdering(nodeId)).take(n)
```
EzequielPostan: note that the change on the metric definition affects this part of the code. 
As your metric is not unidirectional, it will be the case that many nodes will be at the same distance. This is not the case with Kademlia's original metric, where all nodes are always at different distance.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+        s"Illegal attempt to add node id with a length different than the this node id."
+      )
+
+    val d = Xor.d(nodeId, this.baseId)
+
+    val bucket: KBucket = buckets.getOrElseUpdate(d, newKBucket)
+
+    bucket.add(nodeId)
+
+    this
```
EzequielPostan, line 43: I would prefer the method to return Unit.
The expression `bucket.add(nodeId)` has a side effect and the type signature of this method suggest that a new updated instance is returned without affecting the original object's state.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+  def remove(nodeId: BitVector): KBuckets = {
+    buckets
+      .get(Xor.d(nodeId, this.baseId))
+      .foreach(
+        bucket =>
+          bucket
+            .find(_ == nodeId)
+            .foreach(record => bucket.remove(record))
+      )
+    this
```
EzequielPostan: same comment I did for `add` method in line 55. I would suggest to return Unit or remove the state change.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+    if (nodeId.length != this.baseId.length)
+      throw new IllegalArgumentException(
+        s"Illegal attempt to add node id with a length different than the this node id."
+      )
+
+    val d = Xor.d(nodeId, this.baseId)
+
+    val bucket: KBucket = buckets.getOrElseUpdate(d, newKBucket)
+
+    bucket.add(nodeId)
```
EzequielPostan: This is not the addition policy in the paper.
KBuckets shouldn't grow more than k elements. Also, a eviction policy is described in the paper and elements inside buckets are sorted by "last time seen" criteria. 
If we decide to follow a particular variant of the protocol where this condition is relaxed, we need to document it. Consider that the restriction on the number of elements in a bucket aids the security of the protocol (e.g. it helps to avoid table flooding)

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -0,0 +1,109 @@
...
+    s"KBuckets(baseId = ${baseId.toHex}): ${iterator.toList.sorted(new XorOrdering(baseId)).map(id => id.toHex).mkString(", ")}"
+  }
+
+  private def newKBucket: KBucket =
+    new CopyOnWriteArraySet[BitVector]().asScala
+}
+
+object KBuckets {
+
+  type KBucket = mutable.Set[BitVector]
```
EzequielPostan: The paper defined kbuckets as sorted list based on "last time seen" criteria.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KMessage.scala
```diff
@@ -0,0 +1,28 @@
+package io.iohk.scalanet.peergroup.kademlia
+
+import java.util.UUID
+
+import scodec.bits.BitVector
+
+sealed trait KMessage[V] {
+  def requestId: UUID
+  def nodeId: BitVector
+  def nodeRecord: V
```
EzequielPostan: Could we use the names `sourceNodeId` and `sourceNodeRecord`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KMessage.scala
```diff
@@ -0,0 +1,28 @@
...
+}
+
+object KMessage {
```
EzequielPostan, line 13: do we plan to implement PING, FIND_NODE and STORE?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -0,0 +1,52 @@
...
+import io.iohk.scalanet.peergroup.{Channel, PeerGroup}
+import monix.execution.Scheduler
+import monix.reactive.Observable
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.Future
+import scala.util.Failure
+
+trait KNetwork[A] {
+  def server: Observable[(Channel[A, KMessage[A]], KRequest[A])]
```
EzequielPostan: is it correct to have the same type parameter for both the `Channel` address and the `KMessage`?
I would think that the Channel address type could be, for example, a `BitVector` representing a node ID.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
EzequielPostan: I think so, I would advise to be careful about the release of resources.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
```diff
@@ -0,0 +1,161 @@
...
+    extends PeerGroup[BitVector, M] {
+
+  private val log = LoggerFactory.getLogger(getClass)
+
+  override def processAddress: BitVector = kRouter.config.nodeId
+
+  override def initialize(): Task[Unit] = {
+    // The protocol defined here requires that a peer
+    // connects then sends it node record as the first message
+    // (it could send earlier messages but they will be ignored)
```
EzequielPostan, line 30: where is this procedure described?
According to the paper, page 7, last paragraph of section 2.3:
To join the network, a node u must have contact to an already participating node w. Then:
- u inserts w in its corresponding bucket.
- u then performs a node lookup for its own ID.
- u then refreshes all its buckets with the triples it gets from the FIND_NODE queries.
- The queries made by u also may update other nodes buckets.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
```diff
@@ -0,0 +1,161 @@
...
+      .mergeMap(
+        channel =>
+          channel.in.collect {
+            case Left(nodeRecord) =>
+              acceptNodeRecord(channel, nodeRecord)
+          }
+      )
+      .subscribe()
+
+    Task.unit
```
EzequielPostan: Two comments:

First, I believe this implementation violates the semantics of the interface.
When a method returns a `Task`, the interpretation I have is that the computation has not started and won't start until I call `runAsync` (or equivalent method), this carries that no side effect should happen before "running" the Task. Is my interpretation wrong?
In this implementation, the underlying peergroup is actually called before I decide to execute the Task.

Another interesting thing is that the underlying peergroup is never initialized. I think this should never actually receive a message if this is correct.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  override def client(to: BitVector): Task[Channel[BitVector, M]] = {
+    val underlyingChannelTask = Task
+      .fromFuture(kRouter.get(to)) // make the underlying kademlia lookup
+      .flatMap { record => // use the lookup's address info to obtain an underlying channel...
+        debug(s"Routing table lookup returns peer $record. Creating new channel.")
+        underlyingPeerGroup.client(InetMultiAddress(new InetSocketAddress(record.ip, record.tcp)))
```
EzequielPostan: if we are using TCP, shouldn't we ask `underlyingPeerGroup` to have `TCPPeerGroup` type?
Also note that Kademlia actually uses UDP.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
```diff
@@ -0,0 +1,161 @@
...
+          debug(s"Routing table lookup failed for peer ${to.toHex}. Raising an error.")
+          Task.raiseError(new ChannelSetupException[BitVector](to, t))
+      }
+
+    for {
+      // after creating the underlying channel attempt to synchronize node records with the peer
+      underlyingChannel <- underlyingChannelTask
+      syn <- underlyingChannel.sendMessage(Left(kRouter.config.nodeRecord))
+      _ <- Task(debug(s"Syn sent to peer ${to.toHex}"))
+      ack <- underlyingChannel.in.headL
```
EzequielPostan, line 64: where is this synchronization idea coming from? and why is it needed?

EzequielPostan (commented):
I still need to review `KRouter` and the test files, but I already left some comments to address.
A suggestion I would add is to define `NodeID` and `Key` types (both, and allow a method `toKey` in `NodeID`) instead of using `BitVector` directly. In that way you could add some extra type safety and the length check could be done inside constructors instead of adding checks along the code. 
I will finish this review tomorrow.

Thank you for the PR @jtownson and my apologies for my delay in the review.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
EzequielPostan: I would say that querying the same node is not exactly a problem. As we would make alpha queries in parallel, that is expected.
However, I do think that this implementation of `lookup` looks strange. To me, the paper defines as stop criteria that the initiator of the lookup does not receive any new closest nodes after querying the first k nodes in `querySet` and, to me, that implies that the control of the iterative calls to `network.findNodes` should be handled by a "thread" where `querySet` is  updated "globally" after the reception of each response.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/XorSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import io.iohk.scalanet.peergroup.kademlia.Generators._
+import io.iohk.scalanet.peergroup.kademlia.Xor._
+import org.scalacheck.Gen.posNum
+import org.scalatest.FlatSpec
+import org.scalatest.Matchers._
+import org.scalatest.prop.GeneratorDrivenPropertyChecks._
+import scodec.bits.BitVector
+
+class XorSpec extends FlatSpec {
```
EzequielPostan, line 11: I would like to ask to add the unidirectionality property (page 4, last paragraph of section 2.1):
- For any key x and distance D > 0, there is exactly one key y such that distance(x,y) = D.

The test could be implemented as follows: forall x, y and z with y != z then d(x, y) != d(x, z) (which fails with the current metric).

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/XorSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+      case (x, y, z) =>
+        d(x, z) <= d(x, y) + d(y, z) shouldBe true
+    }
+  }
+
+  it should "provide the correct maximal distance" in forAll(posNum[Byte]) { bitCount =>
+    val zero = BitVector.low(bitCount)
+    val max = BitVector.high(bitCount)
+
+    d(zero, max) shouldBe bitCount
```
EzequielPostan: with the paper's metric, this could be 
```
d(zero, max) shouldBe BigInt(2).pow(bitCount)
````
Note that `d` will need to return a `BigInt`

File: scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodecSpec.scala [hunk 24]
```diff
@@ -0,0 +1,78 @@
...
+      // and the second half full of rubbish.
+      // codecs should not be fooled by this.
+      val b = codec.encode(t).toArray
+      val newB = ByteBuffer
+        .allocate(b.length * 2)
+        .put(b)
+        .put(randomBytes(b.length))
+      (newB: java.nio.Buffer).position(0)
+
+      codec.decode(newB).right.value shouldBe t
```
EzequielPostan, line 50: Interesting, but this kinda looks like something to add to decco and not here.
At the same time, is this actually the expected behaviour? Could you expand on the reason?

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodecSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+  "BitVectorCodec" should "work" in testCodec[BitVector]
+}
+
+object BitVectorCodecSpec {
+  import io.iohk.decco.BufferInstantiator.global.HeapByteBuffer
+  import org.scalatest.EitherValues._
+  import org.scalatest.Matchers._
+  import org.scalatest.prop.GeneratorDrivenPropertyChecks._
+
+  private def encodeDecodeTest[T](implicit codec: Codec[T], a: Arbitrary[T], eq: Equivalence[T]): Unit = {
```
EzequielPostan, line 31: isn't this already implemented in decco?

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KBucketsSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.util.Random
+
+class KBucketsSpec extends FlatSpec {
+
+  behavior of "KBuckets"
+
+  val kb = new KBuckets(aRandomBitVector())
+
+  they should "retrieve the base node id" in {
```
EzequielPostan, line 17: minor comment, in the paper, no kbucket holds the node's self id.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KRouterSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+      // assert the bootstrap knows about n1 after n1 contacted it
+      n0.get(n1.config.nodeId).futureValue shouldBe n1.config.nodeRecord
+    }
+
+    "should inform the new node of its neighbourhood" in {
+      val (_, n1, n2) = a3NodeNetwork()
+      // n1 and n2 will not know each other
+      // because they bootstrap from n0
+
+      // assert that n1 and n2 now know about each other after
```
EzequielPostan: I don't follow, "after" what event? bootstrapping?

EzequielPostan (changes requested):
I left my comments above. Thank you again @jtownson, let me know if you need anything else from my side.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: I have copied the signatures from the scala collections. Agreed, they are not the most intuitive but I am keeping for consistency in case we want to extend a scala collection at some point.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: True. I couldn't see a reason why not and it made the code a bit cleaner to store it.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KMessage.scala
jtownson: We plan PING and FIND_NODE.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: Pls see the latest version which simplifies this. When I came to write the test app, some of the abstraction was difficult to understand and served little point; so I removed it.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
jtownson: Good point about Task. I've updated.

Re initialization. The current, unwritten convention is that a peer group does not own its underlying peer group, therefore does not initialize it or shut it down. This convention came about when peer groups supported messages of _any_ type. This is no longer true, peer group message types are fixed when the peer group is instantiated and the the type signatures of underlying peer groups are often specialized to their parents' needs (e.g. `Either[Parent.ControlMessage, M]`). This argues, then, that we should write a story which changes the convention, to where `initialize` on a peer group should initialize its underlying peer group (and so until the terminal peer group) and likewise for shutdown. As it stands today, we do not yet have this, so it is assumed that the user code will initialize both parent and child peer groups (and if it doesn't it won't work). Hope that makes sense.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
jtownson: As discussed, it is a solution to the 'reverse lookup' problem: how do I get the kademlia nodeId of a remote node connecting via the underlying peer group.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodecSpec.scala
jtownson: It's not. I didn't want to introduce a new dependency (scodec) into decco just for the BitVector codec. At some point, if this codec is required more widely, it might make sense to introduce a decco-bitvector module or somesuch.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodecSpec.scala
jtownson: The test is checking that, when decoding, a codec does not read beyond the end of the bytes that were encoded. Thinking about it, it might be the case that this check is now redundant, but I think I'll wait for the dust to settle with Marcin's decco refactor to see.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KBucketsSpec.scala
jtownson: Yes, I know. Doing remote lookups, I have to add code to filter out the node's own id (because it is in the kbuckets) but, equally, it is cleaner to return closest node results from the kbuckets knowing the nodes id _is_ there when that id should be part of the result, as that avoids some cumbersome if logic.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KRouterSpec.scala
jtownson: Yes, bootstrapping. I've clarified the comments in the test.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: See also other comment about this below.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: No, we should not! I've rewritten this method and added tests to ensure it does not leak resources.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KPeerGroup.scala
jtownson: As discussed, it's entirely separate from kademlia.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: I've made the querySet global to lookup process and am now updating the querySet with a nodeId _before_ making a request to that node (in previous code, the querySet was updated when the response came back, leading to a second way of having duplicate requests to a single node). I've also tidied up the impl a bit.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/XorSpec.scala
jtownson: Added. Thanks.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -0,0 +1,235 @@
...
+        }
+        .runAsync
+        .recover {
+          case t: Throwable =>
+            info(s"Query to node $knownNode failed due to exception $t.")
+            Seq.empty
+        }
+    }
+
+    def loop(closestNodes: Seq[BitVector]): Future[Seq[NodeRecord[A]]] = {
```
EzequielPostan, line 147: I have the impression that this method is doing something different from the paper's lookup method. 
From the paper, I understand that the recursive calls end when the first K elements in `closestNodes` (sorted by distance) are contained in `querySet`. However, I believe that in this implementation, the method stops when all the nodes in `closestNodes` have been queried (and not just the first K). Do you agree?
Maybe replacing line 150
```
      val alphaClosestToQuery: Seq[BitVector] = closestNodes
```
with 
```
      val alphaClosestToQuery: Seq[BitVector] = closestNodes.take(config.k)
```
would close the difference

EzequielPostan (approved):
The PR looks much better :)
It still needs to correct discrepancies with the paper though. 
In particular, I think that the lookup implementation is not correct, could you review my comment before merging to clarify this?
In any case, I prefer to merge this and create the stories to iterate over the code than updating the PR again.
Thank you for the PR @jtownson, great work!

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: Yes, I think you are right. I have updated to take k closest nodes.

EzequielPostan: I have the impression that the proper semantic should be:
```
val alphaClosestToQuery: Seq[BitVector] = {
        val (firstK, rest) = closestNodes.filterNot(myself).splitAt(k)
        val notQueriedYet = firstK.filterNot(querySet.contains)
        if(notQueriedYet.isEmpty) notQueriedYet
        else (notQueriedYet ++ rest.filterNot(querySet.contains)).take(config.alpha)
      }
```
But, as mentioned, I am fine with merging the PR and addressing this in the next iteration.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodec.scala
```diff
@@ -0,0 +1,11 @@
+package io.iohk.scalanet.peergroup.kademlia
+
+import io.iohk.decco.Codec
+import io.iohk.decco.auto.instances.NativeInstances.ByteArrayCodec
+import scodec.bits.BitVector
+
+object BitVectorCodec {
+
+  implicit def bitVectorInstance(implicit iCodec: Codec[Int], bCodec: Codec[Boolean]): Codec[BitVector] =
```
krcz: Is `bCodec` used?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
krcz: Mutable Scala collections seem to return `Unit` or other value related to insertion:
* `ListBuffer`: https://www.scala-lang.org/api/2.12.4/scala/collection/mutable/ListBuffer.html#insert(n:Int,elems:A*):Unit
* `Set`: https://www.scala-lang.org/api/2.12.4/scala/collection/mutable/Set.html#add(elem:A):Boolean

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -0,0 +1,84 @@
...
+        clientChannel <- peerGroup.client(to.routingAddress)
+        _ <- clientChannel
+          .sendMessage(message)
+          .timeout(requestTimeout)
+          .doOnFinish(closeIfAnError(clientChannel))
+        nodes <- clientChannel.in
+          .collect { case n @ Nodes(_, _, _) => n }
+          .headL
+          .timeout(requestTimeout)
+          .doOnFinish(closeIfAnError(clientChannel))
```
krcz: Instead of closing the channel in case of errors in each instruction or - if there are no errors - closing it in the end, can't you just use `bracket`?
```
peerGroup.client(to.routingAddress).bracket { clientChannel =>
  for {
    ...
 } yield nodes
} { clientChannel => clientChannel.close() }
```

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -0,0 +1,241 @@
...
+      info(
+        s"Initialization complete. ${nodeRecords.size} peers identified " +
+          s"(of which 1 is myself and ${config.knownPeers.size} are preconfigured bootstrap peers)."
+      )
+  }
+
+  // Handle findNodes requests...
+  network.findNodes.foreach {
+    case (findNodesRequest, responseHandler) =>
+      val (uuid, nodeRecord, targetNodeId) =
```
krcz: `val FindNodes(uuid, nodeRecord, targetNodeId) = findNodesRequest` seems to be a nicer option here

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -0,0 +1,241 @@
...
+    lookup(key).flatMap(_ => getLocally(key))
+  }
+
+  private def getLocally(key: BitVector): Future[NodeRecord[A]] = {
+    toFuture(
+      for {
+        nodeId <- kBuckets.closestNodes(key, 1).find(_ == key)
+        record <- nodeRecords.get(nodeId)
+      } yield record,
+      new Exception(s"Target node id ${key.toHex} not loaded into kBuckets.")
```
krcz, line 84: Nitpicking: as I understand, that is a common situation - there is nothing exceptional in the fact that record is not in the local cache; can we just return `Option` here?

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/Generators.scala
```diff
@@ -0,0 +1,76 @@
...
+  ): Gen[(BitVector, BitVector, BitVector)] =
+    for {
+      v1 <- genBitVector(bitLength)
+      v2 <- genBitVector(bitLength)
+      v3 <- genBitVector(bitLength)
+    } yield (v1, v2, v3)
+
+  def genBitVectorExhaustive(
+      bitLength: Int = defaultBitLength
+  ): List[BitVector] = {
```
krcz, line 39: You could just iterate over numbers and convert them to bit vectors, to make it more concise:
`(0L until (1L << bitLength)).map(n => BitVector.fromLong(n, bitLength))`

Not a strong preference though, you can keep your version if you want.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/Xor.scala
```diff
@@ -0,0 +1,11 @@
+package io.iohk.scalanet.peergroup.kademlia
+
+import scodec.bits.BitVector
+
+object Xor {
+
+  def d(a: BitVector, b: BitVector): BigInt = {
+    assert(a.length == b.length)
+    BigInt((a xor b).toBin, 2)
```
krcz: Isn't there a better conversion way than rendering 160-character string? e.g. `BigInt(1, (a xor b).toByteArray)`?

krcz (commented):
I've left some comments. Resolving some of them might be postponed till further phases of Kademlia implementation, but eventually I'd prefer if it were improved.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/Xor.scala
jtownson: There is!

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: Agreed.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: Good call. Thanks.

jtownson: Re comments about Unit on mutable collections. Yes, okay, point taken, I'll update when sorting out the kbuckets implementation in later stories.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/BitVectorCodec.scala
jtownson: no

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: I am more concerned about the readability `get` so won't change but we can always tinker with these things to make as nice as possible.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/Generators.scala
jtownson: yeah, whatever :-)

krcz (approved):

jtownson:
Merged via command line `git merge --squash`

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: Okay, updated.

PR #47, opened 2019-08-15
Update LICENSE and NOTICE files

EzequielPostan:


jtownson (commented):
The only problem I see @EzequielPostan is that fields on line 190 of the licence file need filling with IOHK's details.

EzequielPostan:
Thank you @jtownson, note that the text you remark is the text added in the NOTICE file.

krcz (approved):

PR #48, opened 2019-08-16
Add UDP tests

EzequielPostan:
These tests explore the interaction between scalanet and an external library under UDP.
With respect to multiplexing, this seems to work as expected. We are discussing possible addressing issues.


jmendiola222:
@EzequielPostan do you think you'll be working on this or should we? Or shall we close the PR? Thanks.

PR #52, opened 2019-09-11
Implement kademlia PING RPC.

jtownson:
This is the implementation of the PING RPC. It is a pre-requisite for getting some of the time-dependent functionality working in kademlia, such as detecting nodes that have gone down.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
@@ -48,7 +48,7 @@ object KNetwork {
     }
 
     override def findNodes: Observable[(FindNodes[A], Nodes[A] => Task[Unit])] = {
-      peerGroup.server().collectChannelCreated.mapTask { channel: Channel[A, KMessage[A]] =>
+      peerGroup.server().collectChannelCreated.mapEval { channel: Channel[A, KMessage[A]] =>
         channel.in
           .collect {
             case f @ FindNodes(_, _, _) =>
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
@@ -99,10 +99,11 @@ object KNetwork {
       } yield response
     }
 
-    private def closeIfAnError(
+    private def closeTheChannel[Request <: KRequest[A]](
         channel: Channel[A, KMessage[A]]
-    )(maybeError: Option[Throwable]): Task[Unit] = {
-      maybeError.fold(Task.unit)(_ => channel.close())
+    )(error: Throwable): (Request, KMessage[A] => Task[Unit]) = {
+      channel.close()
+      null
     }
 
     private def sendResponse(
@@ -117,14 +118,18 @@ object KNetwork {
     private def serverTemplate[Request <: KRequest[A], Response <: KResponse[A]](
         pf: PartialFunction[KMessage[A], Request]
     ): Observable[(Request, KMessage[A] => Task[Unit])] = {
-      peerGroup.server().collectChannelCreated.mapEval { channel: Channel[A, KMessage[A]] =>
-        channel.in
-          .collect(pf)
-          .map(request => (request, sendResponse(channel)))
-          .headL
-          .timeout(requestTimeout)
-          .doOnFinish(closeIfAnError(channel))
-      }
+      peerGroup
+        .server()
+        .collectChannelCreated
+        .mapEval { channel: Channel[A, KMessage[A]] =>
+          channel.in
+            .collect(pf)
+            .map(request => (request, sendResponse(channel)))
+            .headL
+            .timeout(requestTimeout)
+            .onErrorHandle(closeTheChannel(channel))
+        }
+        .filterNot(_ == null)
     }
   }
 }
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
@@ -12,12 +12,15 @@ import monix.reactive.Observable
 trait KNetwork[A] {
 
   /**
-    * Server side FIND_NODES handler.
-    * @return An Observable for receiving FIND_NODES requests.
-    *         Each element contains a tuple consisting of a FIND_NODES request
-    *         with a function for accepting the required NODES response.
+    * Server side requests stream.
+    * @return An Observable for receiving FIND_NODES and PING requests.
+    *         Each element contains a tuple consisting of a request
+    *         with a function for accepting the required response.
+    *         With current conventions, it is mandatory to provide
+    *         Some(response) or None for all request types, in order that the
+    *         implementation can close the channel.
     */
-  def findNodes: Observable[(FindNodes[A], Nodes[A] => Task[Unit])]
+  def kRequests: Observable[(KRequest[A], Option[KResponse[A]] => Task[Unit])]
 
   /**
     * Send a FIND_NODES message to another peer.
@@ -26,14 +29,6 @@ trait KNetwork[A] {
     * @return the future response
     */
   def findNodes(to: NodeRecord[A], request: FindNodes[A]): Task[Nodes[A]]
-
-  /**
-    * Server side PING handler.
-    * @return An Observable for receiving PING requests.
-    *         Each element contains a tuple consisting of a PING request
-    *         with a function for accepting the require PONG response.
-    */
-  def ping: Observable[(Ping[A], Pong[A] => Task[Unit])]
 
   /**
     * Send a PING message to another peer.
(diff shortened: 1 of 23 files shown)


scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -25,6 +26,22 @@ trait KNetwork[A] {
...
     */
   def findNodes(to: NodeRecord[A], request: FindNodes[A]): Task[Nodes[A]]
+
+  /**
+    * Server side PING handler.
+    * @return An Observable for receiving PING requests.
+    *         Each element contains a tuple consisting of a PING request
+    *         with a function for accepting the require PONG response.
+    */
+  def ping: Observable[(Ping[A], Pong[A] => Task[Unit])]
```
krcz: Overloading is good for creating methods doing conceptually the same thing but for different types of parameters. I don't like using such mechanism to introduce two different things (method to send ping and stream of incoming pings).

I see that there's the same convention for `findNodes` already, I must have missed it last time.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -76,13 +105,26 @@ object KNetwork {
...
+    private def serverTemplate[Request <: KRequest[A], Response <: KResponse[A]](
+        pf: PartialFunction[KMessage[A], Request]
+    ): Observable[(Request, KMessage[A] => Task[Unit])] = {
+      peerGroup.server().collectChannelCreated.mapTask { channel: Channel[A, KMessage[A]] =>
+        channel.in
+          .collect(pf)
+          .map(request => (request, sendResponse(channel)))
+          .headL
+          .timeout(requestTimeout)
+          .doOnFinish(closeIfAnError(channel))
```
krcz: I understand that only one message is to be received over such channel, is that correct?

What if there is no message of given type? Does `headL` end with error, closing `channel` in `doOnFinish`? Won't it result in a race condition if you attach two different `serverTemplate`s to one stream?

krcz (changes requested):

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
EzequielPostan: from `timeout` documentation
```
Returns a Task that mirrors the source Task but that triggers a TimeoutException in case the given duration passes without the task emitting any item.
```
I agree about the race condition issue if the same message is "accepted" by `pf` after `collect`.

I believe that if the underlying channel does not deliver a message that passes the filter provided by `collect(pf)` before `requestTimeout` elapses, the channel will be closed and requests will be lost. This can happen if the underlying peergrou is TCP for example, where channels are received when connections start but probably would not happen (if the `requestTimeout` is big enough) in UDP where channels are created when the first message arrive.

krcz: That's right, I was a little tired and I haven't noticed that in such case `headL` itself won't finish until the channel is closed.

I have another question though: how `mapTask` is affected with `Task` that returns error? Doesn't it stop processing the server stream?

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KNetworkSpec.scala
```diff
@@ -13,131 +13,158 @@ import monix.reactive.Observable
...
+      when(channel.in).thenReturn(Observable.eval(request))
+      when(channel.close()).thenReturn(Task.unit)
+
+      val actualRequest = requestExtractor(network).runAsync.futureValue
+
+      actualRequest shouldBe request
+      verify(channel, never()).close()
+    }
+
+    s"Server $label" should "close server channels when a request does not arrive before a timeout" in {
```
EzequielPostan, line 74: this is the scenario I mentioned above. Can a node define a timeout to receive requests? I could understand the timeout for responses, but could one control the incoming request events?

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KNetworkSpec.scala
```diff
@@ -13,131 +13,158 @@ import monix.reactive.Observable
...
+      val (network, peerGroup) = createKNetwork
+      val client = mock[Channel[String, KMessage[String]]]
+      val exception = new Exception("failed")
+      when(peerGroup.client(targetRecord.routingAddress))
+        .thenReturn(Task.raiseError(exception))
+      when(client.close()).thenReturn(Task.unit)
+
+      val t: Throwable = clientRpc(network).failed.evaluated
+
+      t shouldBe exception
```
EzequielPostan, line 161: I see in line 134 `when(client.close()).thenReturn(Task.unit)` 
should you check that `channel.close()` was called or not called at the end of this test?

EzequielPostan (changes requested):
I left some extra comments on the tests. I agree with @krcz's comments and they need to be clarified/addressed.
In general, it looks good. Thank you for the PR @jtownson  :)

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -37,52 +54,82 @@ object KNetwork {
...
+        .collectChannelCreated
+        .mapEval { channel: Channel[A, KMessage[A]] =>
+          channel.in
+            .collect(pf)
+            .map(request => (request, sendResponse(channel)))
+            .headL
+            .timeout(requestTimeout)
+            .onErrorHandle(closeTheChannel(channel))
+        }
+        .filterNot(_ == null)
```
krcz: I can understand the temptation of passing `null` where pair type is expected in order to convey errors, but could we do it in a more idiomatic way, such as `Option`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -37,52 +54,82 @@ object KNetwork {
...
+      peerGroup
+        .server()
+        .collectChannelCreated
+        .mapEval { channel: Channel[A, KMessage[A]] =>
+          channel.in
+            .collect(pf)
+            .map(request => (request, sendResponse(channel)))
+            .headL
+            .timeout(requestTimeout)
+            .onErrorHandle(closeTheChannel(channel))
```
krcz: Let's think how this part works:
* if `pf` is defined at ("collects") the value, it is passed to `headL` and then as task result,
* otherwise `headL` task never completes, so `timeout` fires, but only after some time.

For `serverTemplate` accepting `Ping` messages, and given stream of one message connections, this is going to have following result:
* `Ping` (process the message)
* `FindNodes` (wait for timeout)
* `FindNodes` (wait for timeout)
* `Ping` (process the message)
* `FindNodes` (wait for timeout)
* `Ping' (process the message)

All non-matching messages severely hinder the processing of matching messages, which is a problem, especially when the `requestTimeout` is high.

In my opinion processing `peerGroup.server()` stream independently many times (one for each message type) is a bad design decision and causes more problems than it solves: each of these needs to handle timeouts, causing them to be handled many times. I'd rather have one stream of mixed messages, forcing the user to do dispatching themselves.

scalanet/src/io/iohk/scalanet/reactive/ObservableOps.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+      * Split an observable by a predicate, placing values for which the predicate returns true
+      * to the right (and values for which the predicate returns false to the left).
+      * This is consistent with the convention adopted by Either.cond.
+      */
+    def split(
+        p: T => Boolean
+    )(implicit scheduler: Scheduler): (Observable[T], Observable[T]) = {
+      splitEither[T, T](elem => Either.cond(p(elem), elem, elem))
```
krcz, line 18: I think that just `(o.filter(p), o.filterNot(p))` would be simpler and more efficient, as it doesn't involve creating new object before filtering.

krcz (changes requested):

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: Moreover, even if the consumer only wanted to handle one type of request (ping, say), it would still be necessary to close the channel for other request types (findNodes).

The only way I can see to do that would be for the calling code to pass some kind of message filter at construction time and the knetwork implementation to close the channel for everything not matching the filter.

That functionality is not required by KRouter (it wants to process everything), therefore I agree, a single message stream is a better approach.

jtownson: I've updated the code to provide a single request stream, as you suggest. This is still a little problematic because if the caller naively invokes `requests.filter(...)` or `requests.collect(...)` he introduces a resource leak. I've added a new class, `KNetworkExtension`, which (hopefully) provides methods that filter correctly.

PR #54, opened 2019-10-10
A-824/kademlia dynamic

jtownson:
This PR adds the simplest possible support for bounding server resources and supporting network churn.


scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KRouterSpec.scala
```diff
@@ -42,78 +55,224 @@ class KRouterSpec extends FreeSpec {
...
+
+      // We subsequently set k=3 (to activate eviction logic) and simulate that nodes are unresponsive if their ids
+      // are in either {0100} or {1000, 1001, 1010, 1011, 1100}.
+      // It should then be the case that the unresponsive nodes are evicted and the bucket state corresponds to
+      // bucket 0: (id=0001, d=1)
+      // bucket 1: (id=0010, d=2), (id=0011, d=3)
+      // bucket 2: (id=0101, d=5), (id=0110, d=6), (id=0111, d=7)
+      // bucket 3: (id=1101, d=13), (id=1110, d=14), (id=1111, d=15)
+
+      // We subsequently simulate that the above unresponsive nodes are now responsive (whilst still holding k = 3).
```
mineme0110, line 133: Is this comment by mistake or not required

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -1,45 +1,40 @@
...
   * @param baseId the nodes own id.
   */
-class KBuckets(val baseId: BitVector) {
-
-  private val nodeIds =
-    new ConcurrentSkipListSet[BitVector](new XorOrdering(baseId)).asScala
+class KBuckets(val baseId: BitVector, val clock: Clock) {
 
-  add(baseId)
+  private val buckets = (0 until baseId.length.toInt).map(_ => TimeSet[BitVector](clock)).toList
```
krcz: `List.fill(baseId.length.toInt)(TimeSet[BitVector](clock))` is the idiomatic way: https://www.scala-lang.org/api/2.12.0/scala/collection/immutable/List$.html#fill[A](n:Int)(elem:=%3EA):CC[A]

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -49,16 +44,19 @@ class KBuckets(val baseId: BitVector) {
     * @return true if present
     */
   def contains(nodeId: BitVector): Boolean = {
-    nodeIds.contains(nodeId)
+    if (nodeId == baseId)
```
krcz: Could be also `nodeId == baseId || bucket(nodeId).contains(nodeId)`

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -68,10 +66,33 @@ class KBuckets(val baseId: BitVector) {
...
+    s"KBuckets(baseId = ${baseId.toHex}):\n\t${buckets.indices.map(i => s"$i: ${bucketToString(buckets(i))}").mkString("\n\t")}"
+  }
+
+  private def bucketToString(bucket: TimeSet[BitVector]): String = {
+    s"${bucket.iterator.map(id => s"(id=${id.toBin}, d=${Xor.d(id, baseId)})").mkString(", ")}"
+  }
+
+  def iBucket(b: BigInt): Int = {
+    import BigIntExtentions._
+    b.log2.toInt
```
krcz: As said before, such implementation of log2 is lossy and requires many operations - what you really need here is just `b.bitLength - 1` (which works for any `b` but `0` where it returns `-1`).

`((BigInt(1) << 159) - 1).log2.toInt == 159` (incorrect)
`((BigInt(1) << 159) - 1).bitLength - 1 == 158` (correct)

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -1,45 +1,40 @@
...
 
   /**
     * Find the n nodes closest to nodeId in kBuckets.
     * Return the resulting node records sorted by distance from the nodeId.
     * The indices into the kBuckets are defined by their distance from referenceNodeId.
     */
   def closestNodes(nodeId: BitVector, n: Int): List[BitVector] = {
-    // replace with routing tree described in kademlia paper...
-    nodeIds.toList.sorted(new XorOrdering(nodeId)).take(n)
+    iterator.toList.sorted(new XorOrdering(nodeId)).take(n)
```
krcz: Is it the optimal implementation? If not, could you add a TODO comment that it can be improved in the future?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +49,74 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+    val bucket = kBuckets.bucket(iBucket)
+    debug(s"Handling potential addition of candidate (${nodeRecord.id.toHex}, $nodeRecord) to ibucket $iBucket.")
+    debug(s"iBucket($iBucket) = $bucket")
+    if (bucket.size < config.k) {
+      bucket.add(nodeRecord.id)
+      nodeRecords.put(nodeRecord.id, nodeRecord)
+    } else {
+      // ping the bucket's least recently seen node (i.e. the one at the head) to see what to do
+      val recordToPing = nodeRecords(bucket.head)
+      network.ping(recordToPing, Ping(uuidSource(), config.nodeRecord)).runToFuture.onComplete {
```
krcz: 1. Note: the part running inside of `onComplete` is no longer synchronized. This is important, as `TimeSet` is not thread safe.
2. Even if you add synchronization inside, then two `add` operations might choose the same `bucket.head`, leading to incorrect results, such as attempting to remove it twice (which is not a problem) and adding two new  elements to the bucket (which is a problem, as bucket is growing.

The solution here might be using `Await` here, but that would mean `add` function would synchronously wait for ping result, which might take long time, blocking the operation that called `add`. Another would be keeping global queue / per-bucket queues of elements that were requested to be added and processing them in an actor-like manner.

I'll take a better look tomorrow, analysing interplay between different parts of code.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/BigIntExtentions.scala
```diff
@@ -0,0 +1,40 @@
...
+        }
+      }
+    }
+
+    def log(base: Int): Double = {
+      bi.ln / Math.log(base)
+    }
+
+    def log2: Double = {
+      log(base = 2)
```
krcz: Following implementation of logarithm is 1. quite costly, as it involves many conditional statements and double operations, 2. losing fidelity, as you are operating on doubles. And we might not need such a thing - see the comment in place where it is used.

krcz (changes requested):
Just a first look at the code, I'll need few more I'm afraid, as the ticket is huge.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
```diff
@@ -0,0 +1,65 @@
...
+  */
+object KNetworkRequestProcessing {
+
+  implicit class KNetworkExtension[A](kNetwork: KNetwork[A])(implicit scheduler: Scheduler) {
+
+    type KRequestT = (KRequest[A], Option[KResponse[A]] => Task[Unit])
+    type FindNodesT = (FindNodes[A], Option[Nodes[A]] => Task[Unit])
+    type PingT = (Ping[A], Option[Pong[A]] => Task[Unit])
+
+    def findNodesRequests(): Observable[FindNodesT] = ignoringPing(kNetwork.kRequests).collect {
```
krcz: I remember asking question here in the previous PR, but I never finished that review, so it stayed pending; here it is:

That implementation won't stand when new message is added, unless you convert `ignoringPing` into `ingoringEverythingButFindNodesRequests`. Moreover implementation utilizing `kRequestCatamorph` is quite complex, it took me 5 minutes to understand what's going on (but I'm quite tired, so that might be it).

Simpler version:
```
kNetwork.kRequests.map {
  case (f @ FindNodes(_, _, _), h) => Some((f, h))
  case (_, h) => h(None); None
}.collect { case Some(v) => v }
```

scalanet/src/io/iohk/scalanet/reactive/ObservableOps.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+      * Split an observable by a predicate, placing values for which the predicate returns true
+      * to the right (and values for which the predicate returns false to the left).
+      * This is consistent with the convention adopted by Either.cond.
+      */
+    def split(
+        p: T => Boolean
+    )(implicit scheduler: Scheduler): (Observable[T], Observable[T]) = {
+      splitEither[T, T](elem => Either.cond(p(elem), elem, elem))
```
krcz: Again: why not just `(o.filterNot(p), o.filter(p))` which doesn't involve creating intermediate `Either`s?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
krcz: Also there is a one place in code where it is run as `kBuckets.closestNodes(key, 1).find(_ == key)`, could you at least optimize it there?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
krcz: Thinking about this a little more making it blocking would slow down processing requests. I believe we need a queue of elements to be added, processed as a stream (in an actor-like way). We can schedule a call to talk this over if you want.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/KRouterSpec.scala
jtownson: It's deliberate.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: Yes, that's a bit nicer. Updated.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: True. Updated.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: Yes, of course! That's a much nicer implementation. Updated.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: I've made the implementation more efficient by sorting a single kbucket and then sorting additional buckets if extra results are needed.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: I've (1) updated TimeSet and KBuckets to make them persistent data structures. The router, which remains mutable, then has an AtomicReference to the latest KBucket state.  (2) rechecked the bucket size condition inside the `onComplete`. Unless I've overlooked something, this should then give threadsafe access to the buckets and consistent reads of the bucket state.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/BigIntExtentions.scala
jtownson: Deleted this file.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
jtownson: Sure. I was obsessing about not duplicating the case statement (hence the catamorphism) but I admit it was a bugger to understand. Updated following your suggestion.

scalanet/src/io/iohk/scalanet/reactive/ObservableOps.scala
jtownson: Well, an intermediate Either... so what?!

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
jtownson: This also addresses your later comment about  `nodeId <- kBuckets.closestNodes(key, 1).find(_ == key)` since the value of `n=1` means the `closestNodes` search there will usually access/sort only one bucket.

scalanet/src/io/iohk/scalanet/monix_subject/ConnectableSubject.scala
```diff
@@ -30,5 +30,33 @@ object ConnectableSubject {
   def apply[T]()(implicit s: Scheduler): ConnectableSubject[T] =
     new ConnectableSubject(PublishSubject(), PublishSubject())
   def apply[T](source: Observable[T])(implicit s: Scheduler): ConnectableObservable[T] =
-    ConnectableObservable.cacheUntilConnect[T, T](source, PublishSubject())
+    cacheUntilConnect[T, T](source, PublishSubject())
+
+  // [JMT] Copied and pasted from monix to fix NPE.
```
EzequielPostan, line 35: is this fixed in monix? could we use a version that has this fixed?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/TimeSet.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def apply[T](elems: T*): TimeSet[T] = {
+    addAll(new TimeSet[T](), elems: _*)
+  }
+
+  def apply[T](clock: Clock, elems: T*): TimeSet[T] = {
+    addAll(new TimeSet[T](clock), elems: _*)
+  }
+
+  private def addAll[T](ts: TimeSet[T], elems: T*): TimeSet[T] = {
```
EzequielPostan: it seems that this method could be implemented as `ts.addAll(elens:_*)`
Also, you could replace the uses of this definition (and avoid defining it) by just calling `addAll` on the `ts` parameter.

scalanet/test/src/io/iohk/scalanet/peergroup/kademlia/TimeSetSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+      val ts2 = ssShuffled.foldLeft(ts) { (acc, next) =>
+        val millis = clock.millis()
+        when(clock.millis()).thenReturn(millis + 1)
+        acc.touch(next)
+      }
+
+      ts2.zip(ssShuffled).foreach {
+        case (l, r) =>
+          l shouldBe r
+      }
```
EzequielPostan, line 34: I would suggest to also check that `ts2` and `ssshuffled` have the same length here.
This would detect a failure of refactoring `touch` improperly.

scalanet/src/io/iohk/scalanet/reactive/ObservableOps.scala
EzequielPostan: Some questions:
1. Where is `split` used? IntelliJ only points to the tests.
2. What should happen if I subscribe to one of the returned observables and not to the other? Should the elements in the unsubscribed observable get "lost" before subscription? 
3. It I provide an observable that caches items, will the two output Observables cache them too? How could I free that cache?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
```diff
@@ -0,0 +1,53 @@
...
+            None
+        }
+        .collect { case Some(v) => v }
+
+    private def ignore(
+        handler: Option[KResponse[A]] => Task[Unit]
+    ): Option[KResponse[A]] => Task[Unit] = {
+      handler(None).runToFuture
+      _ => Task.unit
+    }
```
EzequielPostan: if you only care about the effect of `handler(None).runToFuture` and do nothing with the returned values (lines 30 and 41), then why don't you just return `Unit` instead of `Option[KResponse[A]] => Task[Unit]` here?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -1,46 +1,66 @@
...
     * Find the n nodes closest to nodeId in kBuckets.
     * Return the resulting node records sorted by distance from the nodeId.
     * The indices into the kBuckets are defined by their distance from referenceNodeId.
     */
   def closestNodes(nodeId: BitVector, n: Int): List[BitVector] = {
-    // replace with routing tree described in kademlia paper...
-    nodeIds.toList.sorted(new XorOrdering(nodeId)).take(n)
+    val ordering = new XorOrdering(nodeId)
+
+    def loop(nodes: List[BitVector], iLeft: Int, iRight: Int): List[BitVector] = {
```
EzequielPostan: please add tailrec annotation

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -1,46 +1,66 @@
...
-
-    this
+    if (nodeId != baseId) {
+      if (nodeId.length != this.baseId.length)
+        throw new IllegalArgumentException(
+          s"Illegal attempt to add node id with a length different than the this node id."
+        )
+      else {
+        val (iBucket, bucket) = getBucket(nodeId)
+        new KBuckets(baseId, clock, buckets.patch(iBucket, List(bucket + nodeId), 1))
```
EzequielPostan: minor note, you could use `updated` instead of `patch` 

```
new KBuckets(baseId, clock, buckets.updated(iBucket, bucket + nodeId)
```

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -67,11 +87,35 @@ class KBuckets(val baseId: BitVector) {
...
-  def remove(nodeId: BitVector): Boolean = {
-    nodeIds.remove(nodeId)
+  def remove(nodeId: BitVector): KBuckets = {
+    if (nodeId == baseId)
+      throw new UnsupportedOperationException("Cannot remove the baseId")
+    else if (!contains(nodeId)) {
+      this
+    } else {
+      val (iBucket, bucket) = getBucket(nodeId)
+      new KBuckets(baseId, clock, buckets.patch(iBucket, List(bucket - nodeId), 1))
```
EzequielPostan, line 135: minor note: same as with `add`, you could use `updated`

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -39,60 +49,78 @@ object KNetwork {
...
+        .collectChannelCreated
+        .mapEval { channel: Channel[A, KMessage[A]] =>
+          channel.in.refCount
+            .collect { case r: KRequest[A] => r }
+            .map(request => (request, sendOptionalResponse(channel)))
+            .headL
+            .timeout(requestTimeout)
+            .onErrorHandle(closeTheChannel(channel))
+        }
+        .filterNot(_ == null)
```
EzequielPostan: same as mentioned below, why the need of this null?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
```diff
@@ -39,60 +49,78 @@ object KNetwork {
...
     }
 
-    private def closeIfAnError(
+    private def closeTheChannel[Request <: KRequest[A]](
         channel: Channel[A, KMessage[A]]
-    )(maybeError: Option[Throwable]): Task[Unit] = {
-      maybeError.fold(Task.unit)(_ => channel.close())
+    )(error: Throwable): (Request, Option[KMessage[A]] => Task[Unit]) = {
+      channel.close()
+      null
```
EzequielPostan: what is the reason to return null here?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
```diff
@@ -0,0 +1,53 @@
...
+          case (f @ FindNodes(_, _, _), h) =>
+            Some((f, h))
+          case (_, h) =>
+            ignore(h)
+            None
+        }
+        .collect { case Some(v) => v }
+
+    def pingRequests(): Observable[PingT] =
+      kNetwork.kRequests
```
EzequielPostan: in line 25, the call to `findNodesRequests` is consuming the same observable `kNetwork.kRequests`. 
If you will close the channel when you call `ignore`, there may be a race condition here.
Imagine the following snippet:
```
val pings = knetwork.pingRequests().foreach { ... }
val finds = knetwork.findNodesRequests.foreach {...}
```
Aren't both values are consuming `kNetwork.kRequests` stream? if the first `pings` val receives a FindNodes request before the `finds` val then the handler will be called with None (as `ignored` is called). 
In such a case I think it may cause responses to be missed due to closed channels. 
Is the above correct?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -146,7 +194,6 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
             info(s"Query to node $knownNode failed due to exception $t.")
             Seq.empty
         }
-
     }
 
     def loop(closestNodes: Seq[BitVector]): Future[Seq[NodeRecord[A]]] = {
```
EzequielPostan, line 231: It seems that this method did not fix a https://github.com/input-output-hk/scalanet/pull/43#discussion_r314784477 from the previous PR

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+  * If a user of KNetwork wanted to consume only one kind of request,
+  * it is not sufficient to collect or filter the request stream, since it is
+  * still necessary to close channels for excluded request types.
+  * The code to do this is demonstrated here.
+  */
+object KNetworkRequestProcessing {
+
+  implicit class KNetworkExtension[A](kNetwork: KNetwork[A])(implicit scheduler: Scheduler) {
```
EzequielPostan: is this intended to be used anywhere outside tests?

EzequielPostan (changes requested):
I still need to re-check the KNetworkSpec and KRouterSpec files, but I can share some comments/questions.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: What's the reason _not_ to return `null`? 

(The arguments about `null` vs, say `Option`, ignoring whether or not they are valid, refer to pubic defs because, it is claimed, it increases type safety. Within the context of an implementation, we are quite free to eschew types if we wish).

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: See comment below.

scalanet/src/io/iohk/scalanet/monix_subject/ConnectableSubject.scala
jtownson: Apparently not, as of release 3.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/TimeSet.scala
jtownson: Yes, good spot. I must have got distracted while making a change. I've refactored so there's a single addAll method in TimeSet (and removed the redundant one from the companion object).

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
jtownson: Re your comments about `ObservableOps`, I've deleted this troubled file.

Re `KNetworkProcessing`:
Firstly, this is only required by the tests, so I've moved it to the test source tree. Agree with the point about return type and have updated (slightly differently to your suggestion, take a look).

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
jtownson: Yes, it's not entirely clear that these two calls (`pingRequests` and `findNodes`) are mutually exclusive. In any case, I've moved this file to the test sources. It is only used from the tests and this should hopefully remove any concerns that this stuff will find its way into production code.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
loverdos: I guess the idea is to not use `null` at all?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -96,20 +96,24 @@ class KRouter[A](
...
         case Success(_) =>
           // if it does respond, it is moved to the tail and the other node record discarded.
-          bucket.touch(recordToPing.id)
-          debug(s"Ping to ${recordToPing.id.toHex} in bucket $iBucket successful.")
-          info(
-            s"Moving ${recordToPing.id} to head of bucket $iBucket. Discarding (${nodeRecord.id.toHex}, $nodeRecord) as routing table candidate."
-          )
-          debug(s"iBucket($iBucket) = $bucket")
+          if (bucket.size >= config.k) {
+            bucket.touch(recordToPing.id)
```
EzequielPostan: note two things:
1. If I am correct, we don't want that `bucket.size > config.k`, if two methods call `add` concurrently I believe, as @krcz pointed above, that you could add more than k elements as `touch` adds the element when it is not present.
2. note that `touch` is a pure method. So, this call is creating a new updated TimeSet that is not stored anywhere. Hence, the effect is "lost"

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -96,20 +96,24 @@ class KRouter[A](
...
+          }
 
         case Failure(_) =>
           // if that node fails to respond, it is evicted from the bucket and the other node inserted (at the tail)
-          this.synchronized {
-            removeNodeRecord(recordToPing)
-            addNodeRecord(nodeRecord)
-            debug(s"Ping to ${recordToPing.id.toHex} in bucket $iBucket failed.")
-            info(s"Replacing ${recordToPing.id.toHex} with new entry (${nodeRecord.id.toHex}, $nodeRecord).")
+          if (bucket.size >= config.k) {
```
EzequielPostan: I think that you would need this condition inside the `synchronised` block.
Also, the condition may need to be `bucket.size < config.k || (bucket.size == config.k && bucket.contains(recordToPing)) ` as you don't want to increase the number of elements beyond k.

EzequielPostan (commented):
Thank you for the update @jtownson 
I added new comments

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
EzequielPostan: > I guess the idea is to not use null at all?

I do endorse that. Furthermore, I will not approve any PR that adds a `null` reference that can be removed (and this one can).
 
> What's the reason not to return null?

If you don't use the type system, you force everyone to pay attention to your method's implementation and keep track in their head whether they should expect a null reference. Hence, even you super hidden private methods now need to be checked and memorised as safe vs unsafe when used... I will never get why people prefer to do the work in their mind that the compiler can do for free. 

> ... refer to pubic defs...

could you share any reference stating this?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -1,46 +1,66 @@
...
 
   /**
     * Find the n nodes closest to nodeId in kBuckets.
     * Return the resulting node records sorted by distance from the nodeId.
     * The indices into the kBuckets are defined by their distance from referenceNodeId.
     */
   def closestNodes(nodeId: BitVector, n: Int): List[BitVector] = {
-    // replace with routing tree described in kademlia paper...
-    nodeIds.toList.sorted(new XorOrdering(nodeId)).take(n)
+    val ordering = new XorOrdering(nodeId)
```
krcz, line 23: This implementation is not correct:
```
def bv(s: String) = BitVector.fromBin(s).get
val kbuckets = new KBuckets(bv("00000"), Clock.systemUTC).add(bv("00001")).add(bv("00010")).add(bv("00100")).add(bv("01000")).add(bv("10000"))
kbuckets.closestNodes(bv("00111"), 3).map(_.toBin)
```
Returns `List(00100, 00010, 00000)`, should be `List(00100, 00010, 00001)`.

Your bucket ordering is incorrect, you first have to take the bucket `nodeId` falls into, then look into lower subbuckets and only then check higher ones.

Here is the pseudocode of correct ordering (of buckets; non-tailrec):
```
def sortBuckets(nodeId: BitVector) = {
  // loop(pos) returns sorted list of buckets with id >= p
  // ordering has following property: if one bucket appears before the next, all of nodes inside the former are closer to nodeid than ones in the latter
  def loop(pos: Int) = {
    if (p == buckets.size) {
      Nil
    } else {
      if (nodeId(pos) == baseId(pos)) {
        // buckets returned by loop(p + 1) share (p + 1)-bit prefix with baseId, so their nodes xorred with nodeId share (p + 1)-bit prefix with (nodeId xor baseId)  - with the (p + 1)-th byte being 0
        // bucket pos share p-bit prefix and differs on (p + 1) bit, so its elements xorred with nodeId give higher metrics value than ones from loop(p + 1), as they share p-bit prefixes and differ on (p + 1)-th bit, which is 0 for elements from loop(pos + 1) and 1 from p-th bucket
        loop(pos + 1) ++ List(buckets(p))
      } else {
        // elements from buckets(p) and from loop(p + 1) buckets xorred with nodeId share p-bit prefix and differ on (p + 1)-th bit, with elements from buckets(p) giving 0 (as that is the first place where they mismatch with baseId and baseId(p) != nodeId(p)) and ones from buckets returned by loop(p + 1) having 1 - so the former are closer in regards to XOR metric
        buckets(p) :: loop(pos + 1)
      }
    }
  }
}
```

We can have a chat on that tomorrow or I can write the code or you can use unoptimized implementation and leave optimization for next tickets - all options are fine by me.

krcz (changes requested):

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetwork.scala
jtownson: I've used `None`.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+            addNodeRecord(nodeRecord)
+            debug(s"Ping to ${recordToPing.id.toHex} in bucket $iBucket failed.")
+            info(s"Replacing ${recordToPing.id.toHex} with new entry (${nodeRecord.id.toHex}, $nodeRecord).")
+            Continue
+        }
+    }
+  }
+
+  private def addNodeRecord(nodeRecord: NodeRecord[A]): Unit = {
+    nodeRecordIndex.set(NodeRecordIndex(kBuckets.add(nodeRecord.id), nodeRecords + (nodeRecord.id -> nodeRecord)))
```
krcz: That is not an atomic update if that's what you meant. Use `getAndUpdate`

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+  }
+
+  private def removeNodeRecord(nodeRecord: NodeRecord[A]): Unit = {
+    nodeRecordIndex.set(NodeRecordIndex(kBuckets.remove(nodeRecord.id), nodeRecords - nodeRecord.id))
```
krcz: Use `getAndUpdate` if you want an atomic update

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+  }
+
+  private def touchNodeRecord(nodeRecord: NodeRecord[A]): Unit = {
+    nodeRecordIndex.set(NodeRecordIndex(kBuckets.touch(nodeRecord.id), nodeRecords))
```
krcz: Use `getAndUpdate` if you want an atomic update

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+          info(
+            s"Moving ${recordToPing.id} to head of bucket $iBucket. Discarding (${nodeRecord.id.toHex}, $nodeRecord) as routing table candidate."
+          )
+          debug(s"iBucket($iBucket) = $bucket")
+          Continue
+        }
+        .recover {
+          case _ =>
+            // if that node fails to respond, it is evicted from the bucket and the other node inserted (at the tail)
+            assert(bucket.size < config.k)
```
krcz: Why do you expect it to be lower if before pinging it wasn't?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+    loop(nodeRecords.toList)
+  }
+
+  private def doAdd: NodeRecord[A] => Future[Ack] = { nodeRecord =>
+    val (iBucket, bucket) = kBuckets.getBucket(nodeRecord.id)
+    debug(s"Handling potential addition of candidate (${nodeRecord.id.toHex}, $nodeRecord) to ibucket $iBucket.")
+    debug(s"iBucket($iBucket) = $bucket")
+    if (bucket.size < config.k) {
+      addNodeRecord(nodeRecord)
+      Continue
```
krcz, line 116: I don't know how why it compiles but it does - but I'm not sure and I'm wondering if that might be some unsound, is there any implicit conversion happening here? @EzequielPostan do you have an idea?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+        case Failure(t) =>
+          log.info(
+            s"Pong response ${response.requestId} to $to failed with exception: $t"
+          )
+        case _ =>
+      }
+  }
+  
+  private def addAll(nodeRecords: Set[NodeRecord[A]]): Future[Ack] = {
+    def loop(records: List[NodeRecord[A]]): Future[Ack] = records match {
```
krcz, line 99: Nitpicking: `Future.sequence` might be useful in such cases, but loop is perfectly fine solution for that problem.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
+      case Nil =>
+        Continue
```
krcz, line 101: Same as below: how does `Continue` gets converted into `Future[Ack]`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -42,37 +55,102 @@ class KRouter[A](val config: Config[A], val network: KNetwork[A])(
...
   def get(key: BitVector): Future[NodeRecord[A]] = {
     debug(s"get(${key.toHex})")
     getLocally(key).recoverWith { case _ => getRemotely(key) }.recoverWith { case t => giveUp(key, t) }
   }
 
-  def add(nodeRecord: NodeRecord[A]): Unit = {
-    kBuckets.add(nodeRecord.id)
-    nodeRecords.put(nodeRecord.id, nodeRecord)
-    debug(s"Added record (${nodeRecord.id.toHex}, $nodeRecord) to the routing table.")
+  def add(nodeRecord: NodeRecord[A]): Ack = {
```
krcz, line 79: That is blocking, isn't it? You can keep it that way, but please add  comment - that this method won't return until ping-and-add procedure is finished.

krcz (changes requested):

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
EzequielPostan: monix magic. They defined:

```
monix.execution.Ack

object Continue: Ack.Continue.type extends Ack
```
which looks normal, but then exploring the type of Ack we get
```
monix.execution
abstract sealed class Ack extends Future[Ack]
 with Serializable
```

So, Continue extends Ack and Ack extends Future[Ack]

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
EzequielPostan: I understand that this is not blocking (as Ack extends Future[Ack]). However, wasn't there a convention that we should not call `onNext` until the Future returns? Or this only applies to certain overloaded methods that explicitly return `Future[Ack]`?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KNetworkRequestProcessing.scala
EzequielPostan: well, for now I won't block the PR due to this reason. But this actually should be also fixed in the tests. We wouldn't like to find race conditions that lead to flaky tests.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
krcz: Ack can be used where Future[Ack] is needed, but not the other way (we cannot return Future[Ack] where Ack is required). Because of that we can expect it to return synchronously. But you are correct that it is not blocking, there is a buffer, infinite by default. Disregard my first comment.

EzequielPostan: > Ack can be used where Future[Ack] is needed, but not the other way (we cannot return Future[Ack] where Ack is required)

Silly me, you are right.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
jtownson: Hmm, I don't. Changed it to <=.

krcz (approved):

EzequielPostan (approved):
Thank you for the hard work @jtownson and @krcz !

PR #60, opened 2019-12-23
[PM-854] Implement cleaning up closed channels

KonradStaniec:
Till now created channels were added indefinitly to activeChannels map. This could lead to at least to potential problems:
- Exhaustion of resources i.e oom
- Exhaustion od dynamically selected client addresses

This pr tries to fix up this situation. General rules are:
- Library never deletes channels from activeChannels map or closes channels by iteslf. It is entirely in hand of the user of the library.
- Once closed channel can never be open again. 
- Closed channels are cleared from  activeChannels map, by background thread in frequency configured by user.
- Sending message through closed channel is considered to be user error.
- There is slight difference in handling client/server channels on netty side. Client channels are created per 'connection' , so they could be closed also on netty side. On server side, there is actually one server channel so it cannot be closed as this would be equivalent of closing whole server.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
@@ -29,7 +29,6 @@ import io.netty.channel.socket.nio.NioDatagramChannel
 import monix.eval.Task
 import monix.execution.Scheduler
 import monix.execution.atomic.AtomicBoolean
-import monix.reactive.Observable
 import monix.reactive.observables.ConnectableObservable
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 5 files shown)


scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -34,18 +43,31 @@ import scala.util.control.NonFatal
...
   private val log = LoggerFactory.getLogger(getClass)
+
   val serverSubject = ConnectableSubject[ServerEvent[InetMultiAddress, M]]()
 
   private val workerGroup = new NioEventLoopGroup()
 
-  private val activeChannels = new ConcurrentHashMap[ChannelId, ChannelImpl]().asScala
+  private[peergroup] val activeChannels = new ConcurrentHashMap[ChannelId, ChannelImpl]()
+
+  private def cleanupTask = {
```
jtownson: I think maybe `Scheduler.scheduleWithFixedDelay` would be better. In any case, are you certain that `cleanup.cancel()` actually stops the cleanup thread? I'm not sure, but it seems more likely that `cancel()` will stop the `foreach` callback but not stop the underlying `Observable`. Plus, there's no test to verify that `onComplete` is invoked on the val returned by `Observable.intervalAtFixedRate`
You could tidy this up a bit. There is a `def`, only used in the line below but if somebody did call it from somewhere else, it would create a second cleanup schedule. Maybe

jtownson (approved):
One query about the use of Observable. I notice, also, on my branch an intermittent test failure in `SimplePeerGroupSpec`, which may have been caused by this change. Possibly worth checking that.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: I think you are right, `Scheduler.scheduleWithFixedDelay` will be a better choice here. I did not find satisfactory explanation of `Observable.intervalAtFixedRate` behaviour when cancelled, and scheduler docs are much clearer about it.

PR #61, opened 2020-01-03
[PM-1028] Simple kademlia refresh process

KonradStaniec:
Implements kademlia refresh process in simple version used in other p2p implementation i.e performing lookup for random targets:
https://github.com/ethereum/go-ethereum/blob/fa538ee7ed04b1a5e101938d64805d3fcd0eb697/p2p/discover/table.go#L298
https://github.com/status-im/nim-eth/blob/e6164996c8e1ce1a33c17250935fc985b9120830/eth/p2p/discovery.nim#L284

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
@@ -6,9 +6,11 @@ import java.util.Random
 import scodec.bits.BitVector
 
 /**
-  * Skeletal kbucket implementation.
   *
   * @param baseId the nodes own id.
+  * @param clock clock required to keep track of last usage of id in particular bucket
+  * @param buckets list of buckets, each bucket is sorted form least recently seen node id at head and most recently seen
+  *                at the tail
   */
 class KBuckets private (
     val baseId: BitVector,
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
@@ -1,6 +1,7 @@
 package io.iohk.scalanet.peergroup.kademlia
 
 import java.time.Clock
+import java.util.Random
 
 import scodec.bits.BitVector
 
@@ -9,7 +10,11 @@ import scodec.bits.BitVector
   *
   * @param baseId the nodes own id.
   */
-class KBuckets private (val baseId: BitVector, val clock: Clock, val buckets: IndexedSeq[TimeSet[BitVector]]) {
+class KBuckets private (
+    val baseId: BitVector,
+    val clock: Clock,
+    val buckets: IndexedSeq[TimeSet[BitVector]]
+) {
 
   def this(baseId: BitVector, clock: Clock) =
     this(baseId, clock, IndexedSeq.fill(baseId.length.toInt)(TimeSet[BitVector](clock)))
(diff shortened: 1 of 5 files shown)


scalanet/src/io/iohk/scalanet/peergroup/kademlia/KBuckets.scala
```diff
@@ -9,7 +10,11 @@ import scodec.bits.BitVector
   *
```
jtownson, line 9: The comment that this is a 'skeletal' kbuckets implementation is now wrong.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -20,11 +21,33 @@ class KRouter[A](
...
+    */
+  private def startRefreshCycle(): Task[Unit] = {
+    Observable
+      .intervalWithFixedDelay(config.refreshRate, config.refreshRate)
+      .consumeWith(Consumer.foreachTask { _ =>
+        lookup(KBuckets.generateRandomId(config.nodeRecord.id.length, rnd)).map(_ => ())
+      })
+  }
+
   // TODO[PM-1035]: parallelism should be configured by library user
```
jtownson, line 51: Probably it should be some function of alpha, which is already user provided.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -386,14 +409,20 @@ class KRouter[A](
...
 object KRouter {
+  import scala.concurrent.duration._
   case class Config[A](
       nodeRecord: NodeRecord[A],
       knownPeers: Set[NodeRecord[A]],
       alpha: Int = 3,
       k: Int = 20,
-      serverBufferSize: Int = 2000
+      serverBufferSize: Int = 2000,
+      refreshRate: FiniteDuration = 15.minutes
```
jtownson, line 429: This config is somehow a bit leaky. Suggest we remove the buffer size parameter and try to understand the kademlia _m_ parameter. To date, we support kademlia's k parameter (measuring the probability of a key being found given some assumptions about node stability) and alpha (concurrency parameter).

IMO, the parameter we might want to add now is a node _lifetime_ parameter. I.e. some measure of the time that a node can appear in lookup results _after_ that node has been switched off. I _think_ the original kademlia implementation has a parameter called _m_ (measured in minutes). It seems related, unfortunately my knowledge here is hazy.

In the context of this PR, if the only lookups are those from your refresh process then the expiry time of a key would to be dictated by the frequency of the refresh process. For future PRs, other refresh possibilities exist such as a node republishing its ID to the network. I think we would probably want the same lifetime parameter to apply in both cases (since both are just implementation strategies for dropping expired keys).

TLDR; consider using a more generic parameter name, possibly _m_.

jtownson (approved):
Looks good.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
KonradStaniec: it could be a good idea if will leave some note about this under the task `PM-1035`

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
KonradStaniec: tbh for now i would leave it as `refreshRate` because: 
- i could not find proper description of `m` in original paper
- this name  (or similiar ) is used in other kademlia implementations which gives some familiarity to the user
- original paper mentions only using lookups as refresh process. I do not have nothing against other options like ` node republishing its ID to the network`  but i would rather rename this parameter when/if we will start considering those in future

PR #65, opened 2020-01-23
[PM-1035] Parallelism level can now be configured by the user in Config class

MaximilianoRey94:
The subclass KRouter.Config has now a new variable called "parallelism", set for default in 4. This variable is used in the initialization of KRouter, for indicate the Consumer who handle the incoming request the maximum number of thread it should create for this purpose

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
@@ -48,9 +48,8 @@ class KRouter[A](
       })
   }
 
-  // TODO[PM-1035]: parallelism should be configured by library user
   private val responseTaskConsumer =
-    Consumer.foreachParallelTask[(KRequest[A], Option[KResponse[A]] => Task[Unit])](parallelism = 4) {
+    Consumer.foreachParallelTask[(KRequest[A], Option[KResponse[A]] => Task[Unit])](config.parallelism) {
       case (FindNodes(uuid, nodeRecord, targetNodeId), responseHandler) =>
         debug(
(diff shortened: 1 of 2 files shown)


scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -419,14 +418,16 @@ object KRouter {
     *          also bucket maximum size. In paper mentioned as replication parameter
     * @param serverBufferSize maximum size of server messages buffer
     * @param refreshRate frequency of kademlia refresh procedure
+    * @param parallelism level of parallelism the request will be processed
```
KonradStaniec: I would change this comment to something more infromative for potential developer, maybe something like  `number of concurrent tasks which will handle incoming requests` ? (or similar)

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -419,14 +418,16 @@ object KRouter {
...
     */
   case class Config[A](
       nodeRecord: NodeRecord[A],
       knownPeers: Set[NodeRecord[A]],
       alpha: Int = 3,
       k: Int = 20,
       serverBufferSize: Int = 2000,
-      refreshRate: FiniteDuration = 15.minutes
+      refreshRate: FiniteDuration = 15.minutes,
+      parallelism:Int = 4
```
KonradStaniec: there should be space after `:`. Generally `scalafmt` would check this on CI, but for some reason it does not work well for pr from external repos. You could try running `./scalafmt` command to format it automaticlly. (there is section `Formatting the codebase` in the readome)

PR #66, opened 2020-01-24
[PM-1243] Additional concurrent tests

KonradStaniec:
Add higher level protocol which can work on top of any transport layer.
Use it in new tests which handle concurrent request/respones


jtownson (approved):
Fine. It seems a bit curious that much of the test coverage is over code that is itself under test sources. Even if the request/response impl is a bit hacky, it's still a useful thing that probably belongs in the main part of the library.

PR #67, opened 2020-01-29
Nodes records have now authentication. Adapted to Ethereum Node record scheme.

MaximilianoRey94:
-  NodeRecord subclass have now a sign. KRouter's node's id is a 33 byte public key
- Add package crypto
- Add spongycastle to dependencies
- Add StreamObject file

One of the Kademelia implementation TODO indicates Kademlia Node Records, the message object that binds a Kademlia ID with his direction in the sub-net, should be signed, using the model of Ethereum Node Records. This model specify the secp256k1 encryption system must be used.

In this implementation a node’s Kademlia ID is its encoded secp256k1 public key. With this, we solved the problem of binding a public key with a direction on the web, because if a node changes its public key, it changes its direction.

The node record has now a sign of the rest of the node record information, using SHA-256 to make the hash. When a node receives a node record, verify if the sign is valid, and if it’s not, the node record will not add.

This system is important because without any identification, everyone can send a fake node record binding an ID of another node with an IP that he controls. Later, it is possible to mount a malicious node on the web with this ID, who sends fake information causing the web doesn’t trust in the ID and the original node will be beaned.

Also, without signed a user can create malicious nodes with IDs similar to a target node and monopolizes its connection with the net. But if the ID must be a public key that correspond to a private key, its difficult generate a pair of keys with this properties.

Additionally, a sequence number has been added. This number increment himself everytime the node refresh, and should increment manually if the node shutdown and later is online again with other IP.

If there are two record nodes with the same vector ID, the system only register the node record with the higher sequence number. This prevent a node to register obsolete information.

PROBLEMS:
- Kademlia specification says the Node’s ID has 160 bits. In the implementation, this can be different, because nodes only check that a node ID has the same length of its own ID. So, all nodes of the web must have IDs with the same length, that can be or not of 160 bits. With this change, the length of nodes ID must be of 264 bits.
 
- You can’t change your public key without losing your identity. In the blockchain case this is not very important, because there are no funds associated with this. A node ID is only an identification, and a node can change it without important repercussions. If a node is hacked and its private key is stolen, the node must change his ID.


MaximilianoRey94:
This pull request has been created for error. Sorry

PR #70, opened 2020-02-03
[PM-1282] Replace decco with scodec

KonradStaniec:
Reasons for change:
- scodec has simpler interface for creating codecs
- scodec has much more combinator for codecs creation
- scodec is much more widely used in scala world


jtownson (approved):
Good step forward for the library.

PR #71, opened 2020-02-05
Nodes records have now authentication. Adapted to Ethereum Node record scheme

MaximilianoRey94:
# Description
One of the Kademelia implementation TODO indicates Kademlia Node Records, the message object that binds a Kademlia ID with his direction in the sub-net, should be signed, using the model of Ethereum Node Records. This model specify the secp256k1 encryption system must be used.

This system is important because without any identification, everyone can send a fake node record binding an ID of another node with an IP that he controls. Later, it is possible to mount a malicious node on the web with this ID, who sends fake information causing the web doesn’t trust in the ID and the original node will be beaned.

Also  a user can create malicious nodes with IDs similar to a target node and monopolizes its connection with the net.

# Proposed Solution
In this implementation a node’s Kademlia ID is its encoded secp256k1 public key. With this, we solved the problem of binding a public key with a direction on the web, because if a node changes its public key, it changes its direction.

The node record has now a sign of the rest of the node record information, using SHA-256 to make the hash. When a node receives a node record, verify if the sign is valid, and if it’s not, the node record will not add.

Also, it's now impossible to generate nodes IDs arbitrarily, so is more difficult to monopolizes a node connection how is described before.

Additionally, a sequence number has been added. This number increment himself everytime the node refresh, and should increment manually if the node shutdown and later is online again with other IP. If there are two record nodes with the same vector ID, the system only register the node record with the higher sequence number. This prevent a node to register obsolete information.

# Important Changes Introduced
- NodeRecord subclass have now a sign. KRouter's node's id is a 33 byte public key
- Add package crypto
- Add spongycastle to dependencies
- Add StreamObject file
- A crypto package was added. It contains the functions to generate key pairs, sign and verify a signature
- There is now a object NodeRecord with functions for create signed NodeRecords and verify the signature of a NodeRecord. It required a Codec[A]. 
- Krouter now must receive a Codec[A] and the private key, because internally required verify the discovered NodeRecords and create new NodeRecords with his ID and other sequence number of the original
- Generators have been alterated for generate NodeRecords with a valid signature. Also defaultBitLength have been changed to 264
- Nodes now don't accept NodeRecords with a invalid signature or a sequence number lower than the last register they have about this ID (if it have any register)

## Problems
- Kademlia specification says the Node’s ID has 160 bits. In the implementation, this can be different, because nodes only check that a node ID has the same length of its own ID. So, all nodes of the web must have IDs with the same length, that can be or not of 160 bits. With this change, the length of nodes ID must be of 264 bits.
- You can’t change your public key without losing your identity. In the blockchain case this is not very important, because there are no funds associated with this. A node ID is only an identification, and a node can change it without important repercussions. If a node is hacked and its private key is stolen, the node must change his ID.

# Testing
There are testing added in KRouterSpec that verify:
- A node accepts a NodeRecord with a valid signature
- A node doesn't accept a NodeRecord with a invalid signature.
- If a node receive two NodeRecords with valid signature and different sequence number, he store the record with the higher sequence number


build.sc
```diff
@@ -49,6 +49,7 @@ object scalanet extends ScalaModule with PublishModule {
     ivy"org.scodec::scodec-bits:1.1.12",
     ivy"org.scodec::scodec-core:1.11.4",
     ivy"org.scala-lang.modules::scala-parser-combinators:1.1.2"
+	ivy"com.madgag.spongycastle:core:1.56.0.0"
```
KonradStaniec: spongycastle is actually special fork of bouncycastle designed for android and is no longer supported, it would be better to use bouncy castle version 1.64

scalanet/src/io/iohk/scalanet/codec/ECDSASignatureCodec.scala
```diff
@@ -0,0 +1,36 @@
...
+import io.iohk.scalanet.crypto.ECDSASignature
+import scodec.Attempt.{Failure, Successful}
+import scodec.bits.{BitVector, ByteVector}
+import scodec.{Attempt, Codec, DecodeResult, SizeBound}
+import scodec.codecs.implicits.implicitByteVectorCodec
+
+object ECDSASignatureCodec extends Codec[ECDSASignature] {
+  override def decode(bits: BitVector): Attempt[DecodeResult[ECDSASignature]] = {
+    implicitByteVectorCodec.decode(bits) match{
+      case Failure(f) => Failure(f)
```
KonradStaniec: Here you are failing on first failure, which is exact semantic of flatMap  for attempt. This means you can use for comprehensions and not pattern match on each step. example:
```
val a = Codec[Int]
val b = Codec[String]

val c = a.encode(5).require ++ b.encode("hello").require

val d = for {
  a1 <- a.decode(c)
  a2 <- b.decode(a1.remainder)
} yield (a1.value, a2.value)
```

which results in:
```
a: scodec.Codec[Int] = 32-bit signed integer
b: scodec.Codec[String] = string32(UTF-8)

c: scodec.bits.BitVector = BitVector(104 bits, 0x000000050000000568656c6c6f)

d: scodec.Attempt[(Int, String)] = Successful((5,hello))
```

scalanet/src/io/iohk/scalanet/codec/ECDSASignatureCodec.scala
```diff
@@ -0,0 +1,36 @@
...
+      case Successful(r) => implicitByteVectorCodec.decode(r.remainder) match{
+        case Failure(f) => Failure(f)
+        case Successful(s) => Successful(new DecodeResult[ECDSASignature](ECDSASignature(new BigInteger(r.value.toArray),new BigInteger(s.value.toArray)),s.remainder))
+      }
+    }
+  }
+
+  override def encode(value: ECDSASignature): Attempt[BitVector] = {
+    implicitByteVectorCodec.encode(ByteVector(value.r.toByteArray)) match{
+      case Failure(f) => Failure(f)
```
KonradStaniec: use for comprehension instead of pattern matching on each step

scalanet/src/io/iohk/scalanet/codec/EitherCodec.scala
```diff
@@ -0,0 +1,51 @@
+package io.iohk.scalanet.codec
+
+import java.nio.ByteBuffer
+
+import scodec.Attempt.{Failure, Successful}
+import scodec.bits.BitVector
+import scodec.{Attempt, Codec, DecodeResult, Err, SizeBound}
+
+class EitherCodec[A, B](implicit a: Codec[A], b: Codec[B], byteCodec: Codec[Byte]) extends Codec[Either[A, B]] {
```
KonradStaniec: scodec has a special combinator called `either` in `scodec.codecs.either` which is exactly what you want

scalanet/src/io/iohk/scalanet/codec/KMessageCodec.scala
```diff
@@ -0,0 +1,101 @@
...
+import java.nio.ByteBuffer
+import java.util.UUID
+
+import io.iohk.scalanet.peergroup.kademlia.{KMessage, KRouter}
+import scodec.Attempt.{Failure, Successful}
+import scodec.{Attempt, Codec, DecodeResult, Err, SizeBound}
+import scodec.bits.BitVector
+import scodec.codecs.implicits._
+
+class KMessageCodec[A](codec: Codec[KRouter.NodeRecord[A]])(implicit longCodec:Codec[Long],uuidCodec:Codec[UUID],byteCodec:Codec[Byte],bitVectorCodec:Codec[BitVector]) extends Codec[KMessage[A]]{
```
KonradStaniec: Scodec uses shapeless for automatic deriviation of coproduct codecs. Take a look on object 
`io.iohk.scalanet.codec.DefaultCodecs.KademliaMessages`, it is enough to import this object into scope to derive codecs for `KMessage[A]` if there is codec for type `A`.

scalanet/src/io/iohk/scalanet/codec/NodeRecordCode.scala
```diff
@@ -0,0 +1,51 @@
+package io.iohk.scalanet.codec
+
+import io.iohk.scalanet.peergroup.kademlia.{KMessage, KRouter}
+import scodec.Attempt.{Failure, Successful}
+import scodec.{Attempt, Codec, DecodeResult, SizeBound}
+import scodec.bits.BitVector
+
+class NodeRecordCode[A](implicit codec: Codec[A],longCodec:Codec[Long],bitVectorCodec:Codec[BitVector]) extends Codec[KRouter.NodeRecord[A]] {
+  override def decode(bits: BitVector): Attempt[DecodeResult[KRouter.NodeRecord[A]]] = {
+    longCodec.decode(bits) match{
```
KonradStaniec: 1. Use for-comp instead pattern matchins on each step
2. You can import `scodec.codecs.Implicits._` into scope to derive codec for `NodeRecord[A]` if there is codec for a in scope

scalanet/src/io/iohk/scalanet/codec/SeqCodec.scala
```diff
@@ -0,0 +1,43 @@
+package io.iohk.scalanet.codec
+
+import scodec.bits.BitVector
+import scodec.{Attempt, Codec, DecodeResult, SizeBound}
+
+import scala.annotation.tailrec
+
+class SeqCodec[A](implicit subcodec:Codec[A], intCodec:Codec[Int]) extends Codec[Seq[A]] {
```
KonradStaniec: You can derive seq codec form scodec list codec like:
```
    implicit def seqCoded[A](implicit listCodec: Lazy[Codec[List[A]]]): Codec[Seq[A]] = {
      listCodec.value.xmap(l => l.toSeq, seq => seq.toList)
    }
```

it is done this way in `io.iohk.scalanet.codec.DefaultCodecs.General` there are also codec for different types of adresses

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -23,7 +29,7 @@ class KRouter[A](
     val clock: Clock = Clock.systemUTC(),
     val uuidSource: () => UUID = () => UUID.randomUUID(),
     val rnd: Random = new SecureRandom()
-) {
+)(implicit codec: Codec[A]) {
```
KonradStaniec, line 35: I am not sure why the krouter needs `Codec[A]` ?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -494,15 +532,61 @@ object KRouter {
...
+        .get
+        .toByteArray ++ encodedUUID.array()
+      try {
+        crypto.verify(
+          encoded,
+          n.sign,
+          crypto.decodePublicKey(n.id.toByteArray)
+        )
+      } catch {
+        case e: IllegalArgumentException => false
```
KonradStaniec: Using try-catch for control flow, is not best practice. It would better to use `Try` , `Either` or some thing more fucntional for error handling

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -494,15 +532,61 @@ object KRouter {
...
+        .toOption
+        .get
+        .toByteArray ++ encodedUUID.array()
+      val sign = crypto.ECDSASignature.sign(encoded, keyPair)
+      NodeRecord[A](id, routingAddress, messagingAddress, sec_number, sign)
+    }
+    def verify[A](n: NodeRecord[A])(implicit codec: Codec[A]): Boolean = {
+      val encodedUUID = ByteBuffer.allocate(8)
+      encodedUUID.putLong(0, n.seq)
+      val encoded = n.id.toByteArray ++ codec
```
KonradStaniec: seems like a repetion from `apply` above, also the intet is not clear to me here. Maybe it would be better to move it to separate function ? (also i would try to avoid using direct bytebuffer calls)

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -129,49 +137,68 @@ class KRouter[A](
...
   private def pingAndUpdateState(recordToPing: Option[NodeRecord[A]], nodeRecord: NodeRecord[A]): Task[Unit] = {
     recordToPing match {
       case None => Task.now(())
       case Some(nodeToPing) =>
         for {
-          pingResult <- ping(nodeToPing)
+          nodePingResponse <- ping(nodeToPing)
           _ <- routerState.update { current =>
-            if (pingResult) {
+            if (nodePingResponse.isDefined && nodePingResponse.get.seq >= nodeToPing.seq) {
```
KonradStaniec, line 191: You can use option `exists`here

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -129,49 +137,68 @@ class KRouter[A](
...
       .ping(recToPing, Ping(uuidSource(), config.nodeRecord))
-      .map(_ => true)
-      .onErrorHandle(_ => false)
-
+      .map(x => Some(x.nodeRecord))
+      .onErrorHandle(_ => None)
   }
 
+  private def isFreshResult(previewsRegisterNodeRecord: Option[NodeRecord[A]], nodeRecord: NodeRecord[A]): Boolean =
+    previewsRegisterNodeRecord.isDefined && previewsRegisterNodeRecord.get.seq > nodeRecord.seq
```
KonradStaniec: 1. You can use option `exists` here.
2. I do not folow protocol here. Could leave somewhere docs, about sequence number - what validation should be done on them, how to respond to them, when sequence number is fresh enough, how to generete own sequnce number etc.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
MaximilianoRey94: It needs the Codec[A] for use the function "verify" of NodeRecord. This function needs it to encode the "routingAddress" and "messagingAddress" parameters of NodeRecord.

KonradStaniec:
ENR records and ethereum were implemented in https://github.com/input-output-hk/scalanet/pull/96.

PR #72, opened 2020-03-05
[PM-1284] Implementation of dynamic tls peer group based on libp2p specs

KonradStaniec:
TLS peer group suitable for p2p setting, based on libp2p specs: https://github.com/libp2p/specs/blob/master/tls/tls.md

Notable differences:
- Only support one key type - Secp256k1, and only in Ethereum format i.e public key is 65 bytes, where first byte signalise if key is compressed or not, and last 64 bytes are public uncompressed point on Secp256k1 curve. We are supporting only this format as this is current key used by midnight.
- Do not use protobufs to serialize  extension. It is not critical at this stage. For public key we use simple format as: first 8 bits signalise key type, rest of the data is public key.
- We do not have tls 1.3. On java it is supported only from java 11 which does not play nice with scala yet. There is some native option available in netty, but it would require a little bit experimentation.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
@@ -26,6 +26,10 @@ import scodec.bits.BitVector
 import scala.util.Try
 
 private[scalanet] object CryptoUtils {
+
+  /**
+    * Elliptic Curve Groups(ECDHE) recommended by TLS 1.3
+    */
   sealed abstract class SupportedCurves(val name: String)
   case object Secp256r1 extends SupportedCurves("secp256r1")
   case object Secp384r1 extends SupportedCurves("secp384r1")
@@ -81,7 +85,7 @@ private[scalanet] object CryptoUtils {
     new KeyPair(keyFac.generatePublic(spkiKeySpec), keyFac.generatePrivate(pkcs8KeySpec))
   }
 
-  def getKeyFromBytes(bytes: Array[Byte]) = {
+  def getKeyFromBytes(bytes: Array[Byte]): Try[PublicKey] = Try {
     val ecPoint = curve.getCurve.decodePoint(bytes)
     val spec = new ECParameterSpec(curveParams.getCurve, curveParams.getG, curveParams.getN)
     val pubKeySpec = new ECPublicKeySpec(ecPoint, spec)
--- a/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
@@ -35,14 +35,28 @@ private[scalanet] object CryptoUtils {
   case object Secp384r1 extends SupportedCurves("secp384r1")
   case object Secp521r1 extends SupportedCurves("secp521r1")
 
-  val curveName = "secp256k1"
+  private val curveName = "secp256k1"
 
-  val PROVIDER = new BouncyCastleProvider()
+  type SignatureBytes = Array[Byte]
 
-  val curveParams: X9ECParameters = SECNamedCurves.getByName(curveName)
+  private val usedSignatureScheme = "SHA256withECDSA"
 
-  val curve: ECDomainParameters =
+  private val usedKeyScheme = "EC"
+
+  private val PROVIDER = new BouncyCastleProvider()
+
+  private val curveParams: X9ECParameters = SECNamedCurves.getByName(curveName)
+
+  private val curve: ECDomainParameters =
     new ECDomainParameters(curveParams.getCurve, curveParams.getG, curveParams.getN, curveParams.getH)
+
+  private def getEcdsaSignature: Signature = {
+    Signature.getInstance(usedSignatureScheme, PROVIDER)
+  }
+
+  private def getEcKeyFactory: KeyFactory = {
+    KeyFactory.getInstance(usedKeyScheme, PROVIDER)
+  }
 
   def generateKeyPair(secureRandom: SecureRandom): AsymmetricCipherKeyPair = {
     val generator = new ECKeyPairGenerator
@@ -61,19 +75,19 @@ private[scalanet] object CryptoUtils {
     genEcKeyPair(secureRandom, curveName.name)
   }
 
-  def signEcdsa(data: Array[Byte], privateKey: PrivateKey, secureRandom: SecureRandom): Array[Byte] = {
-    val ecdsaSign = Signature.getInstance("SHA256withECDSA", PROVIDER)
+  def signEcdsa(data: Array[Byte], privateKey: PrivateKey, secureRandom: SecureRandom): SignatureBytes = {
+    val ecdsaSign = getEcdsaSignature
     ecdsaSign.initSign(privateKey, secureRandom)
     ecdsaSign.update(data);
     ecdsaSign.sign();
   }
 
-  def verifyEcdsa(data: Array[Byte], sig: Array[Byte], publicKey: java.security.PublicKey): Boolean =
+  def verifyEcdsa(data: Array[Byte], signature: SignatureBytes, publicKey: java.security.PublicKey): Boolean =
     Try {
-      val ecdsaVerify = Signature.getInstance("SHA256withECDSA", PROVIDER)
+      val ecdsaVerify = getEcdsaSignature
       ecdsaVerify.initVerify(publicKey)
       ecdsaVerify.update(data)
-      ecdsaVerify.verify(sig)
+      ecdsaVerify.verify(signature)
     }.fold(_ => false, result => result)
 
   def convertBcToJceKeyPair(bcKeyPair: AsymmetricCipherKeyPair): KeyPair = {
@@ -81,7 +95,7 @@ private[scalanet] object CryptoUtils {
     val pkcs8KeySpec = new PKCS8EncodedKeySpec(pkcs8Encoded)
     val spkiEncoded = SubjectPublicKeyInfoFactory.createSubjectPublicKeyInfo(bcKeyPair.getPublic).getEncoded()
     val spkiKeySpec = new X509EncodedKeySpec(spkiEncoded)
-    val keyFac = KeyFactory.getInstance("EC", PROVIDER)
+    val keyFac = getEcKeyFactory
     new KeyPair(keyFac.generatePublic(spkiKeySpec), keyFac.generatePrivate(pkcs8KeySpec))
   }
 
@@ -89,7 +103,7 @@ private[scalanet] object CryptoUtils {
     val ecPoint = curve.getCurve.decodePoint(bytes)
     val spec = new ECParameterSpec(curveParams.getCurve, curveParams.getG, curveParams.getN)
     val pubKeySpec = new ECPublicKeySpec(ecPoint, spec)
-    val keyFac = KeyFactory.getInstance("EC", PROVIDER)
+    val keyFac = getEcKeyFactory
     keyFac.generatePublic(pubKeySpec)
   }
 
@@ -120,7 +134,7 @@ private[scalanet] object CryptoUtils {
       certificateBuilder.addExtension(extension.oid, extension.isCritical, extension.value)
     }
 
-    val signer = new JcaContentSignerBuilder("SHA256WITHECDSA").build(connectionKeyPair.getPrivate);
+    val signer = new JcaContentSignerBuilder(usedSignatureScheme).build(connectionKeyPair.getPrivate);
 
     val ca = certificateBuilder.build(signer)
 
--- a/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
@@ -1,0 +1,127 @@
+package io.iohk.scalanet.crypto
+
+import java.math.BigInteger
+import java.security._
+import java.security.cert.X509Certificate
+import java.security.spec.{ECGenParameterSpec, PKCS8EncodedKeySpec, X509EncodedKeySpec}
+import java.util.Date
+
+import io.iohk.scalanet.peergroup.dynamictls.DynamicTLSExtension.Extension
+import org.bouncycastle.asn1.sec.SECNamedCurves
+import org.bouncycastle.asn1.x500.X500Name
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo
+import org.bouncycastle.asn1.x9.X9ECParameters
+import org.bouncycastle.cert.X509v3CertificateBuilder
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter
+import org.bouncycastle.crypto.AsymmetricCipherKeyPair
+import org.bouncycastle.crypto.generators.ECKeyPairGenerator
+import org.bouncycastle.crypto.params.{ECDomainParameters, ECKeyGenerationParameters}
+import org.bouncycastle.crypto.util.{PrivateKeyInfoFactory, SubjectPublicKeyInfoFactory}
+import org.bouncycastle.jce.interfaces.ECPublicKey
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.bouncycastle.jce.spec.{ECParameterSpec, ECPublicKeySpec}
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder
+import scodec.bits.BitVector
+
+import scala.util.Try
+
+private[scalanet] object CryptoUtils {
+  sealed abstract class SupportedCurves(val name: String)
+  case object Secp256r1 extends SupportedCurves("secp256r1")
+  case object Secp384r1 extends SupportedCurves("secp384r1")
+  case object Secp521r1 extends SupportedCurves("secp521r1")
+
+  val curveName = "secp256k1"
+
+  val PROVIDER = new BouncyCastleProvider()
+
+  val curveParams: X9ECParameters = SECNamedCurves.getByName(curveName)
+
+  val curve: ECDomainParameters =
+    new ECDomainParameters(curveParams.getCurve, curveParams.getG, curveParams.getN, curveParams.getH)
+
+  def generateKeyPair(secureRandom: SecureRandom): AsymmetricCipherKeyPair = {
+    val generator = new ECKeyPairGenerator
+    generator.init(new ECKeyGenerationParameters(curve, secureRandom))
+    generator.generateKeyPair()
+  }
+
+  def genEcKeyPair(secureRandom: SecureRandom, curveName: String): KeyPair = {
+    val ecSpec = new ECGenParameterSpec(curveName)
+    val g = KeyPairGenerator.getInstance("EC", PROVIDER)
+    g.initialize(ecSpec, secureRandom)
+    g.generateKeyPair();
+  }
+
+  def genTlsSupportedKeyPair(secureRandom: SecureRandom, curveName: SupportedCurves): KeyPair = {
+    genEcKeyPair(secureRandom, curveName.name)
+  }
+
+  def signEcdsa(data: Array[Byte], privateKey: PrivateKey, secureRandom: SecureRandom): Array[Byte] = {
+    val ecdsaSign = Signature.getInstance("SHA256withECDSA", PROVIDER)
+    ecdsaSign.initSign(privateKey, secureRandom)
+    ecdsaSign.update(data);
+    ecdsaSign.sign();
+  }
+
+  def verifyEcdsa(data: Array[Byte], sig: Array[Byte], publicKey: java.security.PublicKey): Boolean =
+    Try {
+      val ecdsaVerify = Signature.getInstance("SHA256withECDSA", PROVIDER)
+      ecdsaVerify.initVerify(publicKey)
+      ecdsaVerify.update(data)
+      ecdsaVerify.verify(sig)
+    }.fold(_ => false, result => result)
+
+  def convertBcToJceKeyPair(bcKeyPair: AsymmetricCipherKeyPair): KeyPair = {
+    val pkcs8Encoded = PrivateKeyInfoFactory.createPrivateKeyInfo(bcKeyPair.getPrivate).getEncoded()
+    val pkcs8KeySpec = new PKCS8EncodedKeySpec(pkcs8Encoded)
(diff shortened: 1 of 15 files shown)


scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+import scala.util.Try
+
+private[scalanet] object CryptoUtils {
+
+  /**
+    * Elliptic Curve Groups(ECDHE) recommended by TLS 1.3
+    */
+  sealed abstract class SupportedCurves(val name: String)
+  case object Secp256r1 extends SupportedCurves("secp256r1")
+  case object Secp384r1 extends SupportedCurves("secp384r1")
```
pslaski, line 35: do we support Secp384r1 and Secp521r1 ?

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    val g = KeyPairGenerator.getInstance("EC", PROVIDER)
+    g.initialize(ecSpec, secureRandom)
+    g.generateKeyPair();
+  }
+
+  def genTlsSupportedKeyPair(secureRandom: SecureRandom, curveName: SupportedCurves): KeyPair = {
+    genEcKeyPair(secureRandom, curveName.name)
+  }
+
+  def signEcdsa(data: Array[Byte], privateKey: PrivateKey, secureRandom: SecureRandom): Array[Byte] = {
```
pslaski: maybe signature `Array[Byte]` could get an alias ? `Signature` or something like this?

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    val ecdsaSign = Signature.getInstance("SHA256withECDSA", PROVIDER)
+    ecdsaSign.initSign(privateKey, secureRandom)
+    ecdsaSign.update(data);
+    ecdsaSign.sign();
+  }
+
+  def verifyEcdsa(data: Array[Byte], sig: Array[Byte], publicKey: java.security.PublicKey): Boolean =
+    Try {
+      val ecdsaVerify = Signature.getInstance("SHA256withECDSA", PROVIDER)
```
pslaski: `Signature.getInstance("SHA256withECDSA", PROVIDER)` could have helper method

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+      ecdsaVerify.update(data)
+      ecdsaVerify.verify(sig)
+    }.fold(_ => false, result => result)
+
+  def convertBcToJceKeyPair(bcKeyPair: AsymmetricCipherKeyPair): KeyPair = {
+    val pkcs8Encoded = PrivateKeyInfoFactory.createPrivateKeyInfo(bcKeyPair.getPrivate).getEncoded()
+    val pkcs8KeySpec = new PKCS8EncodedKeySpec(pkcs8Encoded)
+    val spkiEncoded = SubjectPublicKeyInfoFactory.createSubjectPublicKeyInfo(bcKeyPair.getPublic).getEncoded()
+    val spkiKeySpec = new X509EncodedKeySpec(spkiEncoded)
+    val keyFac = KeyFactory.getInstance("EC", PROVIDER)
```
pslaski: `KeyFactory.getInstance("EC", PROVIDER)` could have a helper method

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    val sn = new BigInteger(64, random)
+    val owner = new X500Name("CN=" + name);
+    val sub = SubjectPublicKeyInfo.getInstance(connectionKeyPair.getPublic.getEncoded)
+    val certificateBuilder = new X509v3CertificateBuilder(owner, sn, beforeDate, afterDate, owner, sub)
+
+    extensions.foreach { extension =>
+      certificateBuilder.addExtension(extension.oid, extension.isCritical, extension.value)
+    }
+
+    val signer = new JcaContentSignerBuilder("SHA256WITHECDSA").build(connectionKeyPair.getPrivate);
```
pslaski: should it be `SHA256withECDSA` ?

scalanet/src/io/iohk/scalanet/peergroup/InetMultiAddress.scala
```diff
@@ -37,3 +45,9 @@ case class InetMultiAddress(inetSocketAddress: InetSocketAddress) {
     state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
   }
 }
+
+object InetMultiAddress {
+  implicit val adressInst = new Addressable[InetMultiAddress] {
```
pslaski: `adressInst` typo... or maybe `addressableInetMultiAddressInst` ;)

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -28,16 +32,16 @@ import scala.concurrent.duration.{FiniteDuration, _}
   * @param state currently open client channels
   * @tparam M the message type.
   */
-class ReqResponseProtocol[M](
-    group: PeerGroup[InetMultiAddress, MessageEnvelope[M]],
-    state: Ref[Task, Map[ChannelId, Channel[InetMultiAddress, MessageEnvelope[M]]]]
-)(implicit s: Scheduler) {
+class ReqResponseProtocol[A, M](
```
pslaski, line 36: `A` param needs a scaladoc

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -110,44 +114,76 @@ object ReqResponseProtocol {
...
+  case object Tcp extends TransportProtocol {
+    override type AddressingType = InetMultiAddress
+
+    override def getProtocol[M](
+        address: InetSocketAddress
+    )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[InetMultiAddress, M]] = {
+      getTcpReqResponseProtocolClient(address)
+    }
+  }
+  case object DTLS extends TransportProtocol {
```
pslaski: does it mean `Datagram Transport Layer Security` or `Dynamic Transport Layer Security` ? :P

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -110,44 +114,76 @@ object ReqResponseProtocol {
...
       address: InetSocketAddress
-  )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[M]] = {
-    val pg1 = new UDPPeerGroup[MessageEnvelope[M]](UDPPeerGroup.Config(address))
+  )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[PeerInfo, M]] = {
+    val codec = implicitly[Codec[MessageEnvelope[M]]]
+    implicit lazy val framingCodec = new FramingCodec[MessageEnvelope[M]](codec)
+    val rnd = new SecureRandom()
+    val hostkeyPair = CryptoUtils.genEcKeyPair(rnd, Secp256k1.curveName)
+    val pg1 =
+      new DynamicTLSPeerGroup[MessageEnvelope[M]](DynamicTLSPeerGroup.Config(address, Secp256k1, hostkeyPair, rnd).get)
```
pslaski: I'm seeing `get` I'm hating :D

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Date
+
+import io.iohk.scalanet.crypto.CryptoUtils
+import io.iohk.scalanet.peergroup.dynamictls.DynamicTLSExtension.SignedKey
+import scodec.bits.BitVector
+
+import scala.util.{Failure, Success, Try}
+
+object CustomTlsValidator {
+  sealed abstract class CertificateError(msg: String) extends CertificateException(msg)
```
pslaski, line 12: do we need this intermediate class? Why not extend `CertificateException(msg)` by objects? I'm just curious ;)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -0,0 +1,135 @@
...
+  case object WrongNumberOfCertificates extends CertificateError("Number of certificates not equal 1")
+  case object WrongCertificateDate extends CertificateError("Certificate is expired or not yet valid")
+  case object NoCertExtension extends CertificateError("Certificate does not have required extension")
+  case object WrongExtensionFormat extends CertificateError("Extension has invalid format")
+  case object WrongExtensionSignature extends CertificateError("Signature on cert extension is invalid")
+  case object SeverIdNotMatchExpected extends CertificateError("Server id do not match expected")
```
pslaski: typo in name

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -0,0 +1,135 @@
...
+      // Certificate keys have different implementation (sun), to use our other machinery it need to be converted to Bouncy
+      // castle format
+      pubKeyInBcFormat <- CryptoUtils.getBouncyCastlePubKey(
+        certificatePublicKey.getEncoded,
+        certificatePublicKey.getAlgorithm
+      )
+      keyBytes <- CryptoUtils.getEcPublicKey(pubKeyInBcFormat)
+      result <- Try(SignedKey.verifySignature(signedKey, keyBytes))
+    } yield result) match {
+      case Failure(_) => Left(WrongExtensionSignature)
```
pslaski: maybe split it to Failure(_) || Success(false)  and Success(true) ?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSExtension.scala
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/**
+  *  Standard Ethereum EC encoding for Secp256k1 public and private keys i.e 65 bytes public key, where
+  *  first indicated if key is compressed or not, and last 64 bytes are public uncompressed point on Secp256k1 curve
+  */
+case object Secp256k1 extends KeyType {
+  val curveName = "secp256k1"
+  val n = 2
+  implicit val Secp256k1disc = Discriminator[KeyType, Secp256k1.type, Int](n)
```
pslaski: Secp256k1disc -> Secp256k1Disc ;)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSExtension.scala
```diff
@@ -0,0 +1,255 @@
...
+      // key must be one from 5.1.1. Supported Elliptic Curves Extension rfc4492, we only support subset which is also
+      // available in tls 1.3 it will ease up migration in the future
+      val connectionKeyPair = CryptoUtils.genTlsSupportedKeyPair(secureRandom, connectionKeyType)
+
+      // safe to call get, as key is generated by us
+      val connectionKeyPairPublicKeyAsBytes = CryptoUtils.getEcPublicKey(connectionKeyPair.getPublic).get
+
+      // Certificate will be valid for next 100 years, the same value is used in libp2p go implementation
+      val today = LocalDateTime.now(Clock.systemUTC())
+      val beforeDate = Date.from(today.minusMonths(1).toInstant(ZoneOffset.UTC))
```
pslaski: is `libp2p go ` uses also one month before date?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+
+  private val log = LoggerFactory.getLogger(getClass)
+
+  private val sslServerCtx: SslContext = DynamicTLSPeerGroupUtils.buildCustomSSlContext(SSLContextForSever, config)
+
+  private val serverSubject = ConnectableSubject[ServerEvent[PeerInfo, M]]()
+
+  private val workerGroup = new NioEventLoopGroup()
+
+  private val clientBootstrap = new Bootstrap()
```
pslaski, line 59: do we need to make it configurable?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+      clientBootstrap,
+      DynamicTLSPeerGroupUtils.buildCustomSSlContext(SSLContextForClient(to), config),
+      codec
+    ).initialize
+  }
+
+  override def server(): ConnectableObservable[ServerEvent[PeerInfo, M]] = serverSubject
+
+  override def shutdown(): Task[Unit] = {
+    serverSubject.onComplete()
```
pslaski, line 105: I don't know the API but `serverSubject.onComplete()` looks strange

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+      _ <- toTask(serverBind.channel().close())
+      _ <- toTask(workerGroup.shutdownGracefully())
+    } yield ()
+  }
+}
+
+object DynamicTLSPeerGroup {
+  case class PeerInfo(id: BitVector, address: InetMultiAddress)
+  object PeerInfo {
+    implicit val peerInInst = new Addressable[PeerInfo] {
```
pslaski: peerInInst -> common :D `peerInfoAddressable` doesn't look bad

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+      codec: StreamCodec[M]
+  ) extends ChannelInboundHandlerAdapter {
+
+    private val log = LoggerFactory.getLogger(getClass)
+
+    override def channelInactive(channelHandlerContext: ChannelHandlerContext): Unit =
+      messageSubject.onComplete()
+
+    override def channelRead(ctx: ChannelHandlerContext, msg: Any): Unit = {
+      val byteBuf = msg.asInstanceOf[ByteBuf]
```
pslaski: is it safe to do casting before try ?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+      .pipeline()
+      .addLast("ssl", sslHandler) //This needs to be first
+      .addLast(new ChannelInboundHandlerAdapter() {
+        override def userEventTriggered(ctx: ChannelHandlerContext, evt: Any): Unit = {
+          evt match {
+            case e: SslHandshakeCompletionEvent =>
+              val localAddress = InetMultiAddress(ctx.channel().localAddress().asInstanceOf[InetSocketAddress])
+              val remoteAddress = InetMultiAddress(ctx.channel().remoteAddress().asInstanceOf[InetSocketAddress])
+              if (e.isSuccess) {
+                // Hacky way of getting id of incoming peer form ssl engine. Only works becouse we have new engine for
```
pslaski: typo

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -0,0 +1,395 @@
...
+
+    override def in: ConnectableObservable[M] = messageSubject
+
+    override def close(): Task[Unit] = {
+      messageSubject.onComplete()
+      toTask(nettyChannel.close())
+    }
+  }
+
+}
```
pslaski, line 158: can you split somehow this file? There is a lot of class implementation which are hard to read

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
```diff
@@ -0,0 +1,110 @@
...
+      Array[TrustManager] { tm }
+    }
+
+    override def engineInit(keyStore: KeyStore): Unit = {}
+
+    override def engineInit(managerFactoryParameters: ManagerFactoryParameters): Unit = {}
+  }
+
+  sealed trait SSLContextFor
+  case object SSLContextForSever extends SSLContextFor
```
pslaski: typo

scalanet/test/src/io/iohk/scalanet/dynamictls/CustomTlsValidatorSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.scalatest.prop.GeneratorDrivenPropertyChecks.forAll
+import org.scalatest.prop.GeneratorDrivenPropertyChecks._
+import org.scalatest.{FlatSpec, Matchers}
+import scodec.bits.BitVector
+
+class CustomTlsValidatorSpec extends FlatSpec with Matchers {
+  val rnd = new SecureRandom()
+  val today = LocalDateTime.now(Clock.systemUTC())
+  val beforeDate = Date.from(today.minusMonths(1).toInstant(ZoneOffset.UTC))
+  val afterDate = Date.from(today.plusYears(100).toInstant(ZoneOffset.UTC))
```
pslaski: as it looks like some const ways to produce before and after date, maybe it would be good to use some factory to get them ?

pslaski (changes requested):

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
KonradStaniec: we support this curves only for certificate creation, there are not available for external user of the library as the whole `CryptoUtils` object is `private[scalanet]`

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
KonradStaniec: generator does not care about it, but i will extract name to variable, so it will be consistent across calls.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
KonradStaniec: ahh you are right about ambiguity here, i will call it by full name here.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
KonradStaniec: ahh common it is only used in tests were we all generating our own key material. But ok  i will do better, and transform Try into Task, it should be prettier. :)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
KonradStaniec: tbh i just like to have my own root hierachy of errors, also it makes sure that all other potential errors will be defined in this file, and in we ever would have to pattern match on error we will have compiler to help.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSExtension.scala
KonradStaniec: nope, i used it to avoid situation that validation fails becouse someones system clock is skewed.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: you mean expose most of this options to the user of library ? I think that eventually it should be done, i would not do this in this pr as it is about different thing. We should probably expose more options in other peer groups but imo it should be separate task.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: it is monix way to tell all subscribers that the stream have ended.  It means that anyone listenning on server events on `serverSubject` will be informed that there will be no more elements.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: yup ale netty messages read from channels are bytebuffers at this level. Also we are casting ot most general bytebuffer type possible so even if implementation would change we should be safe here.

scalanet/test/src/io/iohk/scalanet/dynamictls/CustomTlsValidatorSpec.scala
KonradStaniec: i think you are right this manual conversion everywhere looks shity. I will add https://github.com/nscala-time/nscala-time library, which supports `Interval` class, and have implemented `toDate` conversions. (unfortunatly `X509Certs` require old java `Date` class as args)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSExtension.scala
KonradStaniec: i will change it to day before date, it should be enough for any weirdness in times.

jtownson (approved):

pslaski (approved):

PR #76, opened 2020-04-14
[PM-1287] Change channel interface

KonradStaniec:
Changes channel interface from:
```
trait Channel[A, M] {
 def in: ConnectableObservable[M]
}
```
to 
```
trait Channel[A, M] {
 def in: ConnectableObservable[ChannelEvent[M]]
}
```

Implementation is as simple as possible ie. one set of channel events for all `PeerGroups`. It make sense as both of our peer groups uses all of those events, and it enables testing of all groups by the same code.

If we would treat scalanet more like generic library probably better and more typesafe design would be to have different set of channel events per peer group (as not all peergroups need to have the same events). Maybe like
```
trait Channel[AddressType, ChannelEvent, Message] {
 def in: ConnectableObservable[Either[ChannelEvent, Message]
}
```
and then in PeerGroup:
```
trait PeerGroup[A, M] {
 type ChannelEvent
 def client(to: A): Task[Channel[A, ChannelEvent, M]]
}

```
But for now it would only complicate things and not add any value for midnight, also we do not know it there will be any more peergroups.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
@@ -4,8 +4,9 @@ import java.io.IOException
 import java.net.{InetSocketAddress, PortUnreachableException}
 import java.util.concurrent.ConcurrentHashMap
 
-import scodec.Codec
+import scodec.{Attempt, Codec}
 import io.iohk.scalanet.monix_subject.ConnectableSubject
+import io.iohk.scalanet.peergroup.Channel.{ChannelEvent, DecodingError, MessageReceived}
 import io.iohk.scalanet.peergroup.ControlEvent.InitializationError
 import io.iohk.scalanet.peergroup.InetPeerGroupUtils.{ChannelId, getChannelId, toTask}
 import io.iohk.scalanet.peergroup.PeerGroup.ServerEvent.ChannelCreated
@@ -65,6 +66,16 @@ class UDPPeerGroup[M](val config: Config, cleanupScheduler: Scheduler = Schedule
     }
   }
 
+  private def handleIncomingMessage(channel: ChannelImpl, datagramPacket: DatagramPacket): Unit = {
+    codec.decodeValue(BitVector(datagramPacket.content().nioBuffer())) match {
+      case Attempt.Successful(msg) =>
+        channel.messageSubject.onNext(MessageReceived(msg))
+      case Attempt.Failure(er) =>
+        log.debug("Message decoding failed due to {}", er)
+        channel.messageSubject.onNext(DecodingError)
+    }
+  }
+
   /**
     * 64 kilobytes is the theoretical maximum size of a complete IP datagram
     * https://stackoverflow.com/questions/9203403/java-datagrampacket-udp-maximum-send-recv-buffer-size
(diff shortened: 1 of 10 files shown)


scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -57,10 +58,11 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
         codec.streamDecode(BitVector(byteBuf.nioBuffer())) match {
```
ntallar, line 59: Given that we have this code inside a try, should we catch and report any errors through `UnexpectedError`?

ntallar (changes requested):
Only a single comment and LGTM!

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: good idea :+1:, i do not expect any error to happen here but who knows really -for example netty have different buffer implementations and no all of them support `byteBuf.nioBuffer()` operations, so if would change buffer impl due to lets performance reasons we will be able at least log it at midnight level.

pslaski (approved):

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -83,7 +94,6 @@ class UDPPeerGroup[M](val config: Config, cleanupScheduler: Scheduler = Schedule
               try {
```
ntallar, line 101: Should we report the case that any error happened here as `UnexpectedError`? Why are this `try`, and the other one on which I made a commentm needed?

KonradStaniec: we can. 

But main reason for them is the `finally` block in which realease bytebuffer, that way buffer will be released no matter what error will happen.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -185,11 +203,17 @@ class UDPPeerGroup[M](val config: Config, cleanupScheduler: Scheduler = Schedule
                 // taking it from map but before server pushes message on to it, `onNext` would be called after `onComplete`
                 // which is breach of observer contract.
                 // It is worth investigating if it is only theoretical possibility or additional synchronization is needed
-                messageE.foreach(m => channel.messageSubject.onNext(m))
+                handleIncomingMessage(channel, datagram)
+              } catch {
+                case NonFatal(e) => handleError(ctx, e)
```
ntallar, line 208: Why are we handling only NonFatal here? And isn't `exceptionCaught` called for this errors as well?

KonradStaniec: imo it makes no sense to catch fatal exception here and passing it further as stream event, the proccess will most probably die anyway. 
Tbh i am not what is netty handler policy in case of such errors.

ntallar (approved):
LGTM!
```
     /(__M__)\
    /, ,   , ,\
   /' ' 'V' ' '\       ARV

```

PR #78, opened 2020-05-20
[PM-2021] Do not add self to table

KonradStaniec:
We should not try to add self to lookup table.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
@@ -139,19 +139,23 @@ class KRouter[A](
 
   def add(nodeRecord: NodeRecord[A]): Task[Unit] = {
     info(s"Handling potential addition of candidate (${nodeRecord.id.toHex}, $nodeRecord)")
-    for {
-      toPing <- routerState.modify { current =>
-        val (_, bucket) = current.kBuckets.getBucket(nodeRecord.id)
-        if (bucket.size < config.k) {
-          (current.addNodeRecord(nodeRecord), None)
-        } else {
-          // the bucket is full, not update it but ping least recently seen node (i.e. the one at the head) to see what to do
-          val nodeToPing = current.nodeRecords(bucket.head)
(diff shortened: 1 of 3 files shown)


pslaski (approved):

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -139,19 +139,23 @@ class KRouter[A](
...
-    for {
-      toPing <- routerState.modify { current =>
-        val (_, bucket) = current.kBuckets.getBucket(nodeRecord.id)
-        if (bucket.size < config.k) {
-          (current.addNodeRecord(nodeRecord), None)
-        } else {
-          // the bucket is full, not update it but ping least recently seen node (i.e. the one at the head) to see what to do
-          val nodeToPing = current.nodeRecords(bucket.head)
-          (current, Some(nodeToPing))
+    if (myself(nodeRecord.id)) {
```
ntallar, line 142: Should we add a test for this case?

ntallar:
>2020-05-28 12:45:27,609 [i.i.s.peergroup.kademlia.KRouter] - abf9f9cf364abe7ae2623f54bb8903d5bfe57e85325fdd747a78fc78718b889c9cdfa45c0ee1c78be5765c49b47c3fb633a040ac82c9f54c92f13bf3b02ce842 Initialization complete. 0 peers identified (of which 1 is myself and 0 are preconfigured bootstrap peers).

Should we update this log to remove the line `of which 1 is myself`?

KonradStaniec:
this log is correct, as `self` is always part of the table from the start, check the method `def getIndex[A](config: Config[A], clock: Clock): NodeRecordIndex[A]` , it init our map with self node entry. It is due to the fact than `kBuckets.closestNodes` always returns `baseId` as part of results (which is self node id), so to stay consistent we need to have this id in the map from the beginning.

ntallar (approved):
LGTM!
```
     /(__M__)\
    /, ,   , ,\
   /' ' 'V' ' '\       ARV

```

PR #80, opened 2020-05-28
[POC] Introduce queues instead of subject

KonradStaniec:
This is poc which changes subjects to queues, main reasons to do so:

- Interfaces would return `Observables` instead of `ConnectableObservables` so sharing observalbe is up to user of library.
- PeerGroups does not need `scheduler`, and all events pushing is up to netty threads in synchronous way. This approach makes possible to:
 a) Channel Queues to be single-producer/multiple-consumer which is more performant and plays nice with netty threading model
 b) In the future we could implement back pressure on tcp level
- we could remover `initialize` method, and make `server` initialize server and start streaming.

This is still raw, but i just want some feedback.


ntallar (commented):
Tbh I'm not sure if I know enough of monix subjects to justfiying going one way or the other, but if we are making a change for performance reasons shouldn't we have any benchmarking that justifies it?

KonradStaniec:
I would say performance is additional improvement, and it is only a conjecture  that it is improved. (if this would be main reason then i totally agree about benchmarking)
The main improvement is that we change incoming interfaces to return `Observables` instead of `ConnectableObservable`, so it is up to user of group to decide it he want to share observable between subscribers.
And another main improvement, is that all incoming operations are happening on netty worker groups instead of external scheduler which makes both remaining peergroups better abstractions (they can manage all their resources.)

aakoshh:
Just a few thoughts before looking at the PR more deeply: 

One problem I see with `Channel.in` being a `ConnectableSubject` is that it's a stateful object and if I look at it in a random context such as [initHandshake](https://github.com/input-output-hk/midnight/blob/develop/node/src/main/scala/io/iohk/ethereum/network/handshaker/HandshakeHandler.scala#L48) I don't know if it has been connected yet or not. For example there 3 different consumers are lined up before `connect()` is called, but there's no way to be sure this hasn't happened before. 

Something that looks more insidious to me is the way `refCount` always returns a new `RefCounteObservable`, and whenever an instance of these unsubscribes (when its last subscription ends) it will also [sever the connection](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/monix_subject/ConnectableSubject.scala#L50) inside the common `ConnectableSubject`, effectively unsubscribing everybody forever. So while it's an innocent looking function we have to make sure we only call it once. Meanwhile, independent subscriptions to `Channel.in` such as the one in [ReqResponseProtocol.subscribeForResponse](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala#L96) are fine because they subscribe to the "output" subject of the connectable. 

My first thought was that maybe the `Channel.in` could be a simple `Observable` resulting from the underlying `.refCount`, but that would remove the ability of lining up multiple consumers like in the handshake and obscure the behaviour that the last unsubscription stops consuming messages from Netty for good. The next thought was that `PeerGroup.client` could return a new `UnconnectedChannel` which exposes an `in: Observable` for setting up subscriptions and has a `.connect(): ConnectedChannel` method which starts the consumption and doesn't terminate it until `.close()` is called. That way at least the compiler would tell us what state we are in.

Haven't thought about a `Queue`, I'll have a look at how that works.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -64,25 +62,24 @@ class UDPPeerGroup[M](val config: Config)(
 
   private def removeChannel(channel: ChannelImpl): Unit = {
     activeChannels.remove(channel.channelId)
-    channel.messageSubject.onComplete()
```
aakoshh, line 67: Is there another way we can signal that the channel is closed? I was looking for a `ChannelEvent` representing the closing of the socket, and so far this was the closest I found.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -187,14 +186,20 @@ class UDPPeerGroup[M](val config: Config)(
       val nettyChannel: NioDatagramChannel,
       localAddress: InetSocketAddress,
       remoteAddress: InetSocketAddress,
-      val messageSubject: ConnectableSubject[ChannelEvent[M]],
+      val messageQueue: ConcurrentQueue[Task, ChannelEvent[M]],
```
aakoshh, line 189: I think this should be a private field, so pushing to it is encapsulated to the `ChannelImpl` itself. At least I was looking up and down for where the messages are published, which could be anywhere as the `messageSubject` was accessed directly.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -187,14 +186,20 @@ class UDPPeerGroup[M](val config: Config)(
...
+    val eventLoopAsScheduler = monix.execution.Scheduler(nettyChannel.eventLoop())
 
     val closePromise: Promise[ChannelImpl] = nettyChannel.eventLoop().newPromise[ChannelImpl]()
 
     val channelId = UdpChannelId(nettyChannel.id(), remoteAddress, localAddress)
 
+    def pushEventOnNettyScheduler(ev: ChannelEvent[M]): Unit = {
+      println("Push messages on")
+      messageQueue.offer(ev).runSyncStep(eventLoopAsScheduler)
+    }
```
aakoshh, line 201: Remove `println`? I think it's unnecessary to state that pushing happens on the netty scheduler unless the caller has a choice of another method to call. 

To be clear, the processing of the events will not happen on the netty event loop, right? 

We have to make sure that this step doesn't block because it would hold up the reading/writing of data across all Netty channels.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -302,11 +310,11 @@ class UDPPeerGroup[M](val config: Config)(
       }
   }
 
-  override def server(): ConnectableObservable[ServerEvent[InetMultiAddress, M]] = serverSubject
+  override def server(): Observable[ServerEvent[InetMultiAddress, M]] =
+    Observable.repeatEvalF(serverQueue.poll).cache
```
aakoshh, line 314: Hm, doesn't `.cache` mean that from the time of the first subscription it caches all elements so that it can replay it to the next subscriber? What's the need for this here? Why cache all server events such as new connections coming in? Won't that result in a memory leak?

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -219,7 +224,10 @@ class UDPPeerGroup[M](val config: Config)(
       }
     }
 
-    override def in: ConnectableObservable[ChannelEvent[M]] = messageSubject
+    //TODO how to finish stream from queue?
+    override def in: Observable[ChannelEvent[M]] = {
+      Observable.repeatEvalF(messageQueue.poll)
```
aakoshh, line 229: To be clear: because the queue is single-producer-multiple-consumer, all subscribers get every message, right?

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -315,6 +323,10 @@ class UDPPeerGroup[M](val config: Config)(
 
 object UDPPeerGroup {
 
+  def getMessageQueue[M]() = {
```
aakoshh, line 326:
```suggestion
  def getMessageQueue[M] = {
```
Since there's no state change by calling this method.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -83,14 +82,14 @@ class ReqResponseProtocol[A, M](
       // sending and subsription are done in parallel to not miss response message by chance
       // alse in case of send failure, any resource will be cleaned
       result <- Task
-        .parMap2(c.sendMessage(messageToSend), subscribeForResponse(c.in, messageToSend.id))(
+        .parMap2(c.sendMessage(messageToSend), subscribeForResponse(c.in.share, messageToSend.id))(
```
aakoshh, line 85: What's the need for `.share` here? If I understand it correctly, `c.in` will return a new `Observable` that polls the underlying message queue, and `.share` further creates an `Observable` on top of that one which multi-casts among its consumers, however there's only one consumer in `subscribeForResponse` so that's moot.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -79,6 +77,8 @@ class DynamicTLSPeerGroup[M](val config: Config)(
 
   private lazy val serverBind: ChannelFuture = serverBootstrap.bind(config.bindAddress)
 
+  // FIXME We could probably delete this funcion as it all peer groups it just start a server and there is already `server` function
+  // it could return Task[Observable[ServerEvent[PeerInfo, M]]]
   override def initialize(): Task[Unit] =
```
aakoshh, line 82: I think it would be even better to remove `initialize()` and `shutdown()` from the `PeerGroup` itself and instead use a [Resource](https://typelevel.org/cats-effect/api/cats/effect/Resource.html) to wrap creation of a peer group instance and encapsulate the cleanup as well:

```scala
object DynamicTLSPeerGroup {
  def apply(config: Config)(
    implicit codec: StreamCodec[M],
    scheduler: Scheduler
  ): Resource[Task, DynamicTLSPeerGroup[M]] = 
  Resource.make(
    for {
      pg <- Task.eval(new DynamicTLSPeerGroup(config))
      _   <- pg.initialize()
    } yield pg
  )(_.shutdown())
}
```

This would make sure that users of the `PeerGroup` trait itself can't just call `.shutdown()` whenever they want.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -113,6 +114,15 @@ class DynamicTLSPeerGroup[M](val config: Config)(
 }
 
 object DynamicTLSPeerGroup {
+  def getMessageQueue[M](typeOf: ChannelType = monix.execution.ChannelType.SPMC) = {
+    ConcurrentQueue.unsafe[Task, M](BufferCapacity.Unbounded(None), typeOf)
+  }
+
+  def pushEventOnNettyScheduler[M](queue: ConcurrentQueue[Task, M], ev: M)(implicit s: Scheduler): Unit = {
```
aakoshh, line 121: I guess this was copy pasted, but obviously the naming here is misleading as the scheduler itself is implicit.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -209,16 +207,18 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
...
-  )(implicit scheduler: Scheduler) {
+  ) {
+
     private val log = LoggerFactory.getLogger(getClass)
     val sslHandler = sslServerCtx.newHandler(nettyChannel.alloc())
 
-    val messageSubject = ConnectableSubject[ChannelEvent[M]]()
     val sslEngine = sslHandler.engine()
+    val channelScheduler = Scheduler(nettyChannel.eventLoop())
+    val channelQueeue = getMessageQueue[ChannelEvent[M]]()
```
aakoshh, line 221:
```suggestion
    val channelQueue = getMessageQueue[ChannelEvent[M]]()
```

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -238,30 +238,31 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
                   s"Ssl Handshake server channel from $localAddress " +
                     s"to $remoteAddress with channel id ${ctx.channel().id} and ssl status ${e.isSuccess}"
                 )
-                val channel = new ServerChannelImpl[M](nettyChannel, peerId, codec, messageSubject)
-                serverSubject.onNext(ChannelCreated(channel))
+                val channel = new ServerChannelImpl[M](nettyChannel, peerId, codec, channelQueeue)
+                pushEventOnNettyScheduler(serverQueue, ChannelCreated(channel))(channelScheduler)
```
aakoshh, line 242: I'm missing something here: why does every instance of `ServerChannelImpl` share the same `channelQueue` (or `messageSubject`)? Doesn't that mean that they all publish into the same common queue/subject? It doesn't seem to make sense then that every channel then had the `in` as a `ConnectableObservable` if it's all just the same one. In fact I don't understand how the handshake protocol works: if it calls `.connect()` on an instance which is common to all channels.

In the `UDPPeerGroup` example each channel has its own subject, independent of the others.

aakoshh: In that light I don't understand how [ReqResponseProtocol.startHandling](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala#L107) didn't have the problems with the multiple `refCount` invocations, that is, every channel created will result in a new `RefCountObservable` being created on top of the same `ConnectableObservable`, which I thought would lead to the first one that gets cancelled unsubscribing everyone else as well.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: Ah, no, SPMC just means that there are multiple consumers vying for the same messages, like a worker queue, so a 2nd subscription to in will actually steal the messages from the 1st, no?

aakoshh:
As per my initial comment, I agree that there could be some changes to the interface to make it easier to reason about the state of the stream of messages, but I think (unless I'm misunderstanding the way it works) creating `Observable`s on top of the `ConcurrentQueue` is not giving us the same semantics as we have now, that is, once connected, every subscription to `in` sees the same data. By the looks of it this could be achieved by using a [ConcurrentChannel](https://monix.io/api/3.0/monix/catnap/ConcurrentChannel.html) instead, but I haven't thought through how.

I'd either expose the message queue directly on the channel, or try to simplify the observables on their own.

aakoshh:
By the looks of it, `ConcurrentChannel` needs consumers set up before messages are pushed to it or they get blackholed. 

I like the idea of exposing the `ConcurrentQueue` as `Channel.in`: it makes it clear that the buffering of messages is happening from the beginning (i.e. `ConnectableSubject` is already consuming the data from the socket, even if we never connect), and it would be up to the user of the `Channel` to decide how they want to consume the data, be it one message at a time or set up an `ConnectableObservable` on top of it and do broadcasting between them. 

It would make the handshake protocol (the consumption of the first 2 messages) trivial as well:
```scala
for {
  first <- channel.in.poll.map(fC.isCompatible(config.first, _)).rethrow
  second <- channel.in.poll.map(sC.isCompatible(config.second, _)).rethrow
  peer = toPeerInfo(first, second)
} yield HandshakedChannel(peer, channel.in)
```
There wouldn't be a need for the complicated dance with the `MVar`. 
This is assuming there's nobody else consuming the queue of course.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: Each PeerGroup has a global `Subject` (or here queue) which provides incoming channels. i.e `channel.in` provides incoming messages and channel events, `peergroup.in` provides incoming channels and peergroup events.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: yea, it should probably take event loop as param to make it clear

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: I am not sure about it to be honest, as a user of library i would like to close my resources as i want. I can of course use bracket and compose it nicely with the rest of the code, but maybe i need to shutdown it along with actorsystem in some random place.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
KonradStaniec: yep you are right, i just wanted here to leave types/scenario as it was without much changes. If we decide that it is worth progressing with that change, then it should be just Observable.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: SPMC here means, that where will be at most one writer thread to that queue, which is guaranted by netty threading model that each channel uses only one thread, and there can be many reader threads, as user can use this observable in many ways and reads its data from many threads.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: yep you are right, i have probably experimented with it and forgot to delete.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: sure, println was only for some debugging purposes only.

Yep processing of events will happen during observable execution on user scheduler. So this method will not block, if the queue is unboudned.

Pushing event on netty thread, makes is possible not to provide external scheduler to peergroup, also it should be more performant as there is no context switch i.e the same thread which received message on netty side pushes it to queue.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: yes, that was also my first thouht, adding new channel event to signal finishg.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -219,7 +224,10 @@ class UDPPeerGroup[M](val config: Config)(
       }
     }
 
-    override def in: ConnectableObservable[ChannelEvent[M]] = messageSubject
+    //TODO how to finish stream from queue?
```
aakoshh, line 227: I think the solution for that would be to:
1. Add a new `case class ChannelClosed(reason: cats.effect.ExitCase) extends ChannelEvent[Nothing]`
2. Instead of `repeatEvalF` use [unfoldEval](https://monix.io/api/current/monix/reactive/Observable$.html#unfoldEval[S,A](seed:=%3ES)(f:S=%3Emonix.eval.Task[Option[(A,S)]]):monix.reactive.Observable[A]), seeded with the queue, which would return `None` after the exit. Or use a custom subscriber that polls the queue in a loop and raises `onError` and `onComplete` appropriately, recursing when the `F[Ack]` returned by `onNext` is completed.

The problem is that if we emit an exit event, there can be strictly one consumer of the queue, or else only one of them gets it. With multiple consumers it would be up to the user to make sure the queue is relayed into a `ConcurrentChannel` (with consumers already set up) or there's one multicasting `Observable`.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -315,6 +323,10 @@ class UDPPeerGroup[M](val config: Config)(
 
 object UDPPeerGroup {
 
+  def getMessageQueue[M]() = {
+    ConcurrentQueue.unsafe[Task, M](BufferCapacity.Unbounded(None), monix.execution.ChannelType.SPMC)
```
aakoshh, line 327: I thought there should be some capacity, to protect ourselves from being DoS'd. Unfortunately that doesn't play very well with the notion of using a channel event to signal the end of processing: if we try to close the channel for any reason, and push an exit event to the queue, if it's full we can't, and we'd either have to:
1. Block by using `queue.push(...)` but we'd be blocking `Channel.close()` which is not what we want.
2. Use `queue.tryPush(...)` but that would fail if the queue is full, so consumers would not be notified, which is bad.
3. Use `queue.push(...).startAndForget` so we can continue closing and also make sure the message eventually gets there and consumers are freed, no resource leaks.

KonradStaniec:
I really like the comments, scalanet lacked this type of discussion for some time :) 

So from the beginning:
1.  I pretty much agree with critic of the ConnectableObservable ( especially `refCount` which can suddenly cancel whole stream when last subscriber unsubscribes). Idea with `UnconnectedChannel` sounds interesting, but i would need to experiment with to understand better how this api would look like. At first glance it sounds a little to complicated i.e for a user of library it can be a little to clear, why he has connected to some peer and received  `UnconnectedChannel`. 

2. I agree that having `Observable` on top of `ConcurrentQueue` does not give us the same semantics, but if such semantics is something user of the library need then he can easy use: `def cacheUntilConnect[A, B](source: Observable[A], subject: Subject[A, B]): ConnectableObservable[B]`. We leave this choice in library user hands. Ultimatly `channel.in` could have only one subcriber through the whole life time of the appliction and then whole `ConnectableObservable` is just unnecessary.
Also my plan for later was, to add back pressure to the queue i.e track queue size, and if the queue size is larger than some max  then disable netty `AUTO_READ`. Then all unread bytes will start accumulating at socket buffer level, and unless queue size will start going down, those buffer will overflow and stop accepting writing from remote peer (something like:
https://github.com/RichRelevance/scalaz-netty/blob/5105e36e29c9f667267b26a20a3cc9b9457a5675/src/main/scala/scalaz/netty/BPAwareQueue.scala)

3. About exposing queue,  I really like the way how handshake would like with also explicit buffering sounds nice, but I see 2 disadvatages of it: 
a) We exposing pretty low level primitive, i would say even implementation detail, and interface level. If we would ever change it that would be pretty breaking. (for example after some time and performance measuremnts we can came to conclusion that we shoud use java `ConcurrentLinkeQueue`)
b) queue has also `push` methods, which user could accidentally use (or on purpose). Then we need to use MPMC ( Multi-producer, multi-consumer) queue and we loss chance for optimization

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: You can do that by calling [Resource.allocated](https://typelevel.org/cats-effect/api/cats/effect/Resource.html#allocated[G[x]%3E:F[x],B%3E:A](implicitF:cats.effect.Bracket[[%CE%B1$7$]G[%CE%B1$7$],Throwable]):G[(B,G[Unit])]) which gives you back the method to call if you want to free your resource explicitly. 

My suggestion is that the ActorSystem and in fact everything that needs closing should be refactored into a `Resource`.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Yes that's how I understand it as well, however:

In `UDPPeerGroup` every `ChannelImpl` that gets created [gets it own](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala#L156) channel event subject, independent of the others.

By contrast in `DynamicTLSPeerGroup` the equivalent is using a [single instance](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala#L212) of a channel event subject that gets [passed](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala#L233) to ever incoming channel, so they all publish into the same subject. I think they should be separate.

KonradStaniec: oh ok i see from where the confusion is coming from. Actually this class should probably be called `ServerChannelPiplieBuilder` as its created for each separate incoming channel:
https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala#L72

So each channel end up with it own builder and its own `messageSubject`

aakoshh:
Good points. 

1. Sorry maybe `ConnectableChannel` is a better name there since everything so far has been `Connectable*`. 
The API would look something like this:
```scala
trait Channel[A, M] {
  def to: A
  def sendMessage(message: M): Task[Unit]
  def in: Observable[ChannelEvent[M]]
  def close(): Task[Unit]
}
/** On this stage you can add subsrcribers to `channel.in` but they won't receive messages just yet because 
  * the underlying `ConnectableObservable` is still caching, allowing you to set up multiple consumers if you need to. */
trait ConnectableChannel[A, M] extends Channel[A, M] {
  /** Trigger to connection of the underlying `ConnectableObservable`. */
  def connect(): ConnectedChannel[A, M] =
}
/** At this stage you can again attach subscribers to `channel.in` but the observable is already hot, they are only 
  * going to receive subsequent messages. */
trait ConnectedChannel[A, M] extends Channel[A, M]
```

2. About backpressure: cool, I wanted to dig into how you can suspend the socket in Netty. We can make use of [Subscriber.request](https://monix.io/api/3.0/monix/execution/rstreams/Subscription.html#cancel():Unit) as well to control exactly when the reading/suspending is triggered without having to deal with overflowing the queue. 

3. I agree that it may be too concrete of an implementation detail to expose, and definitely about the `push` method. Maybe the channel should have the `poll` methods, but that may be going too far. 

If we restricted ourselves to having just a single subscriber to the observable then I'd be happy with the Observable sitting on top of the Queue, I'd just like it if the interface expressed this, to avoid exchanging one footgun for another. The `ConnectableObservable` at least gave total control to the user, so it was obvious that something like this is happening, whereas the simple `Observable` obscures this detail. This is where I start to understand how Rust's ownership model would help: transforming the variable from one state to another would make the original reference invalid, so only one place could use that observable.

aakoshh:
Maybe `channel.in` should be a `Try[Observable[_]]` so it can only be taken once, otherwise return an `IllegalStateException`.

KonradStaniec:
oh I think I understand what case you is worrying you, something like:
```
val channel = PeerGroup.client...
val firstSub = channel.in.foreach(...).runAsyncAndForget
val secondSub = channel.in.foreach(...).runAsyncAndForget
```

I am not sure if we should guard user from such misuse. Docs of `Observable` are pretty clear that observable should not be shared, and if one want to share it thare are methods for it like `.share` or `.cacheUntilConnect`. 

Also maybe `channel.in` should probably be defined as val i.e `val in = Observable.repealEvalF(queue.poll)` thats way it is at least clear that there is only one observable and if user meses something with too many subscribers then it is his fault. In my view the concept of subscribers and cosumers, is a little outside of the library scope. On library level i want to just say to user `you can have unshareable stream of channel events` which Observable convey pretty well.

Also wrapping return type with `Try[Observable[_]]` maybe good indirect detterent fo calling `.in` to often, but maybe makingin it effectful would be better like: `Task[Observable[_]]` , and then in definition:
```
val in = Task {
  Observable.repetEvalF(queue.poll)
}.memoizeOnSucces
```
?

Also i need to think on `channel.poll` as it sound pretty interesting! I mean,i like how it is symmetric with `channel.sendMessage` i.e one pushes one message to remote peer, another asynchoronusly get one event from remote peer. It gives a lot of flexibility, and user can build of data stream as he likes.  Then we could provide a package/class `ChannelStream` or something like that which would provide some typical stream connections. It would be big change but it sounds neat and composable.

aakoshh:
> I am not sure if we should guard user from such misuse. Docs of Observable are pretty clear that observable should not be shared, and if one want to share it thare are methods for it like .share or .cacheUntilConnect

The thing with `Observable` is that it's not clear from looking at it whether it's hot or cold. I agree that the library provides ways to deal with this, it would just be better if we could make misuse glaringly obvious. 

About `Task(Observable.repeatEvalF(queue.poll)).memoizeOnSuccess`: I don't think this works the way I thought I proposed. The returned observable is just a blueprint of how to get data, it doesn't do anything until it's subscribed to, and when it's subscribed, it will start polling, each subscription independently of the other, so it's still load balancing the data among the subscribers:

```scala
@ val q = ConcurrentQueue.unbounded[Task, Int]().runSyncUnsafe(1.second) 
@ val ot = Task(Observable.repeatEvalF(q.poll)).memoizeOnSuccess 
@ val o1 = ot.runSyncUnsafe(1.second).foreach(x => println(s"1 / $x")) 
@ val o2 = ot.runSyncUnsafe(1.second).foreach(x => println(s"2 / $x")) 
@ q.offerMany(List(1,2,3,4,5)).runSyncUnsafe(1.second) 

1 / 1
2 / 2
2 / 3
1 / 4
2 / 5
```

What I meant is that we could make sure there's only one subscriber if, on subsequent calls, we return an error. For example the Observable could be put in an MVar during setup, and taken by the first subscription, with the rest returning an error instead. `channel.in` could be `def in(): Task[Option[Observable]]` or something backed by `_in.tryTake`. 

About `Observable.repeatEvalF(queue.poll)` itself: I commented on your TODO that this wouldn't finish, and it would possibly lead to memory leaks by keeping the subscribers on the queues forever trying to poll, and that pushing a "close event" would possibly be blocked by bounded capacity. Here's another scheme I thought might work:

```scala
class Channel[M] {
  /** Indicate whether the channel was closed and if any of the leftover messages should be consumed. */
  private val closed = cats.effect.concurrent.Deferred.tryable[Task, Bool]()
  
  def close(): Task[Unit] = {
    ...
    closed.complete(false)
  }

  /** Wait for the next message or return None if the channel got closed. */
  def poll(): Task[Option[M]] = {
    closed.tryGet flatMap {
      case Some(false) => 
        Task now None
      case Some(true) =>
        queue.tryPoll()
      case None =>
        Task.racePair(queue.poll(), closed.get) flatMap {
           case Left(m) => Task now Some(m)
           case Right(_) => poll()
        }
    }
  }

  /** Anybody using `in` can be sure they won't miss a message from the queue, once they are subscribed, 
    * regardless of how many other subscriptions there are. But it's a hot Observable, and the messages buffered
    * until the first subscription happens are taken by the first subscriber; subsequent subscribers will only see
    * the messages that arrive after their subscription.
    * If one needs to set up multiple consumers that all receive every message from the very beginning they can do so:
    * `val c = ConnectableObservable.cacheUntilConnect(in, PublishSubject()); c.foreach(...); c.foreach(...); c.connect()` */
  val in: Observable[M] =
    Observable.repeatEvalF(poll()).takeWhile(_.isSome).map(_.get).share
}
```

By the way, have you considered using `fs2.Streams`? It's a bit more obvious that a stream isn't supposed to be consumed by multiple consumers, but it can be combined in such ways I'm sure. It also has a `Queue` to support pushing into it.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Ah, thanks, I knew I was missing something :see_no_evil:

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: Forget what I said here, I think https://github.com/input-output-hk/scalanet/pull/80#issuecomment-677552959 is a better way to signal channel closure and plays nicely with the bounded queues too.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: https://github.com/input-output-hk/scalanet/pull/80#issuecomment-677552959 handles the case of accidental multiple consumers as well, forget the event.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: By the way, if you consider renaming that class, I'd actually turn it into a method, rather than a constructor that is used as a method, with the resulting class immediately discarded, so this:
```scala
      override def initChannel(ch: SocketChannel): Unit = {
        new ServerChannelBuilder[M](serverSubject, ch, sslServerCtx, codec.cleanSlate)
        log.info(s"$processAddress received inbound from ${ch.remoteAddress()}.")
      }
```
would be 
```scala
     override def initChannel(ch: SocketChannel): Unit = {
        ServerChannelPipelineBuilder.build[M](serverSubject, ch, sslServerCtx, codec.cleanSlate)
        log.info(s"$processAddress received inbound from ${ch.remoteAddress()}.")
      }
```
When I first saw it I wasn't sure the instance wasn't accidentally lost.

KonradStaniec:
Summary of current status and probable way forward:

1. We will change subjects to queues and introduce backpresoure on TCP level by netty `AUTO_READ` option
2. We will remove method `channl.in` from channel, and instead add methods in shape `def readEvent(): Task[ChannelEvent[M]]`. Then user of library will be in charge of consuming data from channel.

Great discussion @aakoshh :)

aakoshh:
We did it in other PRs.

PR #81, opened 2020-06-02
[PM-2104] Fix hack in tls group

KonradStaniec:
Fixes a hack in tls peergroup used to pass client peer id to other parts of the code. 

Main gain from it is, that now upgrading to tls 1.3 is just a question of upgrading netty and putting native tls libraries on class path. (without this fix native libraries throwed exception during handshake)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -13,7 +13,7 @@ object scalanet extends ScalaModule with PublishModule {
 
   def scalaVersion = "2.12.10"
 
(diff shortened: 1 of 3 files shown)


pslaski (approved):

ntallar (approved):
LGTM!
```
     /(__M__)\
    /, ,   , ,\
   /' ' 'V' ' '\       ARV

```

PR #84, opened 2020-08-25
PM-[1860,2396]: Use scala-logging

aakoshh:
Instead of using the `slf4j.Logger` interface directly, use [scala-logging](https://github.com/lightbend/scala-logging). The benefit is that if the DEBUG level is not enabled then the macros won't evaluate the arguments.

Wrapped logs in `Task.delay` where it was `Task.now` so that we get referential transparency. `Task.now(logger.info(...))` would emit the log as a side effect of creating the task, but `.delay` suspends it. 

`import cats.syntax.apply._` is to get access to the `task1 *> task2` syntax which is part of monix 3.2.2, but not 3.0.0. Updated to 3.1.0 for now since that's what the downstream project is using, and updating to 3.2.2 would cause `Task.gather` and `Task.wander` to be deprecated. But I think we should upgrade to 3.2.2 to get any bug fixes.

We agreed with @KonradStaniec to do this as part of the tickets mentioned in the title, but the actual fixes they refer to will come separately.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -39,9 +39,10 @@ object scalanet extends ScalaModule with PublishModule {
   )
 
   override def ivyDeps = Agg(
-    ivy"io.monix::monix:3.0.0",
+    ivy"io.monix::monix:3.1.0",
     ivy"com.github.nscala-time::nscala-time:2.22.0",
     ivy"com.chuusai::shapeless:2.3.3",
+    ivy"com.typesafe.scala-logging::scala-logging:3.9.2",
     ivy"org.slf4j:slf4j-api:1.7.25",
     ivy"io.netty:netty-all:4.1.31.Final",
     ivy"org.eclipse.californium:scandium:2.0.0-M15",
--- a/build.sc
+++ b/build.sc
@@ -39,7 +39,7 @@ object scalanet extends ScalaModule with PublishModule {
   )
 
   override def ivyDeps = Agg(
-    ivy"io.monix::monix:3.1.0",
+    ivy"io.monix::monix:3.2.2",
     ivy"com.github.nscala-time::nscala-time:2.22.0",
     ivy"com.chuusai::shapeless:2.3.3",
     ivy"com.typesafe.scala-logging::scala-logging:3.9.2",
@@ -107,7 +107,7 @@ object scalanet extends ScalaModule with PublishModule {
       ivy"com.github.pureconfig::pureconfig:0.11.1",
       ivy"com.github.scopt::scopt:3.7.1",
       ivy"org.scodec::scodec-bits:1.1.6",
-      ivy"io.monix::monix:3.0.0",
+      ivy"io.monix::monix:3.2.2",
       ivy"org.scala-lang.modules::scala-parser-combinators:1.1.2"
     )
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
@@ -236,9 +236,9 @@ class UDPPeerGroup[M](val config: Config)(
 
     private def closeChannel(): Task[Unit] = {
       for {
-        _ <- Task.delay(logger.debug("Closing channel from {} to {}", localAddress: Any, remoteAddress: Any))
+        _ <- Task(logger.debug("Closing channel from {} to {}", localAddress: Any, remoteAddress: Any))
         _ <- closeNettyChannel(channelType)
-        _ <- Task.delay(logger.debug("Channel from {} to {} closed", localAddress: Any, remoteAddress: Any))
+        _ <- Task(logger.debug("Channel from {} to {} closed", localAddress: Any, remoteAddress: Any))
       } yield ()
     }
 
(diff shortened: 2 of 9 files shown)


build.sc
```diff
@@ -39,9 +39,10 @@ object scalanet extends ScalaModule with PublishModule {
   )
 
   override def ivyDeps = Agg(
-    ivy"io.monix::monix:3.0.0",
+    ivy"io.monix::monix:3.1.0",
```
KonradStaniec: As we are updating deps, whay you think about updating also netty as we are using some ancient version ?

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -235,9 +236,9 @@ class UDPPeerGroup[M](val config: Config)(
 
     private def closeChannel(): Task[Unit] = {
       for {
-        _ <- Task.now(log.debug("Closing channel from {} to {}", localAddress: Any, remoteAddress: Any))
+        _ <- Task.delay(logger.debug("Closing channel from {} to {}", localAddress: Any, remoteAddress: Any))
```
KonradStaniec: why not `Task(...)` ? It is a little shorter, and from what i remember task `apply` and `delay`, are same thing

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
```diff
@@ -255,33 +256,33 @@ class UDPPeerGroup[M](val config: Config)(
...
-      log.debug("Sending message {} to peer {}", message, recipient)
-      Task.fromTry(codec.encode(message).toTry).flatMap { encodedMessage =>
-        val asBuffer = encodedMessage.toByteBuffer
-        toTask(nettyChannel.writeAndFlush(new DatagramPacket(Unpooled.wrappedBuffer(asBuffer), recipient, sender)))
-          .onErrorRecoverWith {
-            case _: IOException =>
-              Task.raiseError(new MessageMTUException[InetMultiAddress](to, asBuffer.capacity()))
-          }
-      }
+      Task.delay(logger.debug("Sending message {} to peer {}", message, recipient)) *>
```
KonradStaniec: maybe as now we have chain of 3 flatmaps, maybe it is worth to use for comprehension here ?

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -132,58 +136,51 @@ class KRouter[A](
...
-            if (pingResult) {
-              info(
-                s"Moving ${nodeToPing.id} to head of bucket. Discarding (${nodeRecord.id.toHex}, $nodeRecord) as routing table candidate."
-              )
-            } else {
-              info(s"Replacing ${nodeToPing.id.toHex} with new entry (${nodeRecord.id.toHex}, $nodeRecord).")
-            }
-          }
-        } yield ()
+        ping(nodeToPing).ifM(
```
KonradStaniec, line 177: This change has different semantics from the old code.
Old version was an: `atomic touch or replace based on ping result`, new version is: `atomic touch or atomic replace based on ping result`.  
Tbh i do not remember why it was important to keep it that way, but i would rather avoid semantic changing refactors, and if that part was overly synchronised just create task to improve that.

build.sc
aakoshh: Sure, shall we do it as another PR?

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: Yeah `apply`, `eval` and `delay` are all the same. I got used to writing [Sync[F].delay](https://typelevel.org/cats-effect/api/cats/effect/Sync.html#delay[A](thunk:=%3EA):F[A]) when using cats-effect, that's all. Can change it if you prefer the shorter version.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: Sure, no probs.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
aakoshh: There cannot be any difference between those, the ping result doesn't change based on whether you're checking inside or outside. What I wanted to get rid of is having to check it twice: once for the change, then once for the logging. That way they can easily get out of sync, if you have to repeat the same logic twice. The tests caught a problem when I used it wrong. And I'd rather not log inside `modify`, that just feels wrong.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
KonradStaniec: yeah I would prefer using `apply` consistenlty

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
KonradStaniec: sure I would also not log in `modify`. So lets leave it as it is, I cannot think of reason why this approach would break

aakoshh: Maybe `modify` uses `compareAndSet` under the hood and when logs were together with the mutation inside people saw them repeatedly, when there was a contention, that's why it was moved out.

build.sc
aakoshh: Do you want to update to monix 3.2.2? I guess Midnight depends on a specific release of scalanet, so there won't be any sudden deprecation warnings there, but I don't know who else might use this. I think it would make sense.

KonradStaniec: Midnight depends on scalanet `0.3` which i have released last week, so its safe. From what I talked with Jeremy atala do not use scalanet, so i would say lets go for it.

scalanet/src/io/iohk/scalanet/peergroup/UDPPeerGroup.scala
aakoshh: Done.

KonradStaniec (approved):

build.sc
aakoshh: Sweet, done.

PR #85, opened 2020-08-25
PM-1860: Sanitize peers before logging

aakoshh:
The problem was that the debug report about lookup results contained duplicates because the algorithm preformed the lookup with the closest nodes in parallel and then just concatenated the results, so if multiple peers knew about the same nodes, they were included repeatedly. 

The PR changes the result of the `lookup` function to be a `Set` rather than a `Seq`. We didn't apply any ordering at that point, so `Set` describes the result better. `lookup` is only used internally, it's followed by a read on the modified local caches, so there are no unit tests for it separately, the log was the only indication of the duplicates.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
@@ -175,18 +175,19 @@ class KRouter[A](
     recordToPing match {
       case None => Task.unit
       case Some(nodeToPing) =>
-        ping(nodeToPing).ifM(
-          // if it does respond, it is moved to the tail and the other node record discarded.
-          Task(
-            logger.info(
-              s"Moving ${nodeToPing.id} to head of bucket. Discarding (${nodeRecord.id.toHex}, $nodeRecord) as routing table candidate."
-            )
-          ) *>
-            routerState.update(_.touchNodeRecord(nodeToPing)),
-          // if that node fails to respond, it is evicted from the bucket and the other node inserted (at the tail)
-          Task(logger.info(s"Replacing ${nodeToPing.id.toHex} with new entry (${nodeRecord.id.toHex}, $nodeRecord).")) *>
-            routerState.update(_.replaceNodeRecord(nodeToPing, nodeRecord))
-        )
+        Task(logger.debug(s"Pinging ${nodeToPing.id.toHex} to check if it needs to be replaced.")) *>
+          ping(nodeToPing).ifM(
+            // if it does respond, it is moved to the tail and the other node record discarded.
+            Task(
+              logger.info(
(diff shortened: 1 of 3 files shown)


scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -354,41 +359,40 @@ class KRouter[A](
...
+          closestKnownNodes.toList.map(_.id -> RequestScheduled).toMap
         // All initial nodes are scheduled to request
-        val lookUpTask: Task[Seq[NodeRecord[A]]] = for {
+        val lookUpTask: Task[Set[NodeRecord[A]]] = for {
           // All initial nodes are scheduled to request
           state <- Ref.of[Task, Map[BitVector, RequestResult]](initalRequestState)
           // closestKnownNodes are constrained by alpha, it means there will be at most alpha independent recursive tasks
-          results <- Task.parTraverse(closestKnownNodes) { knownNode =>
-            recLookUp(List(knownNode), NonEmptyList.fromListUnsafe(closestKnownNodes.toList), state)
+          results <- Task.parTraverse(closestKnownNodes.toList) { knownNode =>
```
KonradStaniec: I am thinking, we are constantly doing this dance `.toList`, `.toSet` and`(nodesFound ++ receivedNodes).distinct(xorOrder).sorted(xorOrder)` (in recLookupFunction)
Do you think it would make sense to use `NonEmptySet` ? (from what I rember it uses scala `SortedSet`, se we would have nicly ordered and diststinct results)

aakoshh: That sounds like a great idea!

aakoshh: Done.

scalanet/src/io/iohk/scalanet/peergroup/kademlia/KRouter.scala
```diff
@@ -354,46 +362,50 @@ class KRouter[A](
...
-          results <- Task.parTraverse(closestKnownNodes) { knownNode =>
-            recLookUp(List(knownNode), NonEmptyList.fromListUnsafe(closestKnownNodes.toList), state)
-          }
-          records = results.flatMap(_.toList)
-        } yield records
+          val lookUpTask: Task[SortedSet[NodeRecord[A]]] = for {
+            // All initial nodes are scheduled to request
+            state <- Ref.of[Task, Map[BitVector, RequestResult]](initalRequestState)
+            // closestKnownNodes are constrained by alpha, it means there will be at most alpha independent recursive tasks
+            results <- Task.parTraverse(closestKnownNodes.toList) { knownNode =>
```
KonradStaniec: just noticed that this could probably be: `parTraverseUnordered`, also (https://github.com/input-output-hk/scalanet/pull/85/files#diff-3efccc289265a0426b8121d2294a1b1bR342)

KonradStaniec (approved):
minor comments but in general LGTM.

PR #88, opened 2020-09-29
ETCM-161: Move the kademlia package into its own submodule.

aakoshh:
Moves the current Kademlia packages from the `scalanet` module to a separate one so that we can change it to implement Ethereum Discovery v4 without introducing Ethereum specific dependencies to the core. Changed the package path from `io.iohk.scalanet.peergroup.kademlia` to just `io.iohk.scalanet.kademlia`. 

I left a comment in the README about publishing multiple modules with mill. I'm not sure how it works at the moment or how it's going to realise which modules need to be published (scalanet and kademlia).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,36 +9,36 @@ Scalanet is an asynchronous, strongly typed, resource-managed networking library
 What does all that mean?
  * Resource managed. Scalanet makes it as easy as possible to send and receive messages without having to open or close connections.
  * Asynchronous. Scalanet is non-blocking. In this regard, it is like netty, however, unlike netty, Scalanet uses *reactive*
- programming idioms. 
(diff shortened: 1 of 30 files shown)


KonradStaniec (approved):
LGTM! 

About mill muli module support, i am also not sure how it works right now. I am wondering if we should not migrate to sbt in some time, as i do not like this friction of using 2 different build tool when jumping betwen projects. Wdyt?

aakoshh:
Thanks for the quick review @KonradStaniec!

I don't know, I like to keep an open mind and `mill` seemed refreshing, but not that well documented, bit of a trial and error. It also didn't want to work without manual hacking with Bloop now. We can use `sbt`, I have no preference.

jmendiola222:
@KonradStaniec let's bring up the `sbt` vs `mill` discussion on next VSU. ;)

PR #89, opened 2020-09-30
ETCM-162: Rename UDPPeerGroup to DynamicUDPPeerGroup

aakoshh:
Renames the existing `UDPPeerGroup` to `DynamicUDPPeerGroup` to reflect the fact that it always creates a new client channel from a random port, in contrast to the `StaticUDPPeerGroup` which will reuse the same port (to be developed).

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
@@ -138,13 +138,13 @@ object ReqResponseProtocol {
         address: InetSocketAddress
     )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[AddressingType, M]]
   }
-  case object Udp extends TransportProtocol {
+  case object DynamicUDP extends TransportProtocol {
     override type AddressingType = InetMultiAddress
 
     override def getProtocol[M](
         address: InetSocketAddress
     )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[InetMultiAddress, M]] = {
-      getUdpReqResponseProtocolClient(address)
+      getDynamicUdpReqResponseProtocolClient(address)
     }
   }
 
@@ -173,7 +173,7 @@ object ReqResponseProtocol {
     } yield prot
   }
 
-  def getUdpReqResponseProtocolClient[M](
+  def getDynamicUdpReqResponseProtocolClient[M](
       address: InetSocketAddress
   )(implicit s: Scheduler, c: Codec[M]): Task[ReqResponseProtocol[InetMultiAddress, M]] = {
     val pg1 = new DynamicUDPPeerGroup[MessageEnvelope[M]](DynamicUDPPeerGroup.Config(address))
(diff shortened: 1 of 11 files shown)


scalanet/src/io/iohk/scalanet/peergroup/udp/DynamicUDPPeerGroup.scala
```diff
@@ -30,24 +29,27 @@ import scodec.{Attempt, Codec}
 import scala.util.control.NonFatal
 
 /**
-  * PeerGroup implementation on top of UDP.
+  * PeerGroup implementation on top of UDP that always opens a new channel
+  * from a random port when sending a message to a remote address.
```
KonradStaniec: I find this doc a little bit confusing.  What it describes is the way that KNetwork used this peer group. In theory nothing keeps user from creating channl per peer (each with randomly assigned port) and then using those channels to communicate with remote peer.

I would try to be more specific and say that each call to `PeerGroup.client` generates randomly assigned port to communicate with remote peer.

scalanet/test/src/io/iohk/scalanet/peergroup/UDPPeerGroupSpec.scala
```diff
@@ -12,6 +12,7 @@ import io.iohk.scalanet.peergroup.ControlEvent.InitializationError
 import org.scalatest.concurrent.ScalaFutures._
 import io.iohk.scalanet.peergroup.PeerGroup.{ChannelAlreadyClosedException, MessageMTUException}
 import io.iohk.scalanet.peergroup.StandardTestPack._
+import io.iohk.scalanet.peergroup.udp.DynamicUDPPeerGroup
```
KonradStaniec, line 15: We probably should rename whole spec to `DynamicUDPPeerGroupSpec`

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -175,7 +176,7 @@ object ReqResponseProtocol {
   def getUdpReqResponseProtocolClient[M](
```
KonradStaniec: there is `case object Udp extends TransportProtocol` I would re-name it to `DynamicUDP`

scalanet/src/io/iohk/scalanet/peergroup/udp/DynamicUDPPeerGroup.scala
aakoshh: You are right, that's more precise, thanks :+1:

scalanet/test/src/io/iohk/scalanet/peergroup/UDPPeerGroupSpec.scala
aakoshh: I wasn't sure yet if we can reuse this spec to test both implementations. Seems to delegate to some "standard tests" so I thought this might be the case. If not I'll rename it.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
aakoshh: Sure thing.

scalanet/test/src/io/iohk/scalanet/peergroup/UDPPeerGroupSpec.scala
KonradStaniec: that sounds reasonable, let's leave it as it is then.

KonradStaniec (approved):

PR #90, opened 2020-10-01
ETCM-169: Use Resource to control PeerGroup and Channel lifecycle

aakoshh:
The PR moves `PeerGroup.initialize` and `PeerGroup.shutdown` from the trait to a `Resource` created by `DynamicTLSPeerGroup` and `DynamicUDPPeerGroup` so there can be no doubt that cleanup will happen. I noticed that throughout the tests that just wasn't the case, not consistently. In theory we didn't need to call `initialize` to use for example `client` to open a channel to a remote address, however `shutdown` should always have been called since the way the peer groups were implemented meant that they started Netty event loops as soon as they were created. It also wasn't obvious whether the initialization happened yet, whether `server` will actually yield events or not. I think it's within the spirit of the p2p PeerGroup that the node is actually supposed to be reachable at the `processAddress`, so opening the ports shouldn't be a problem. The incoming requests will be buffered until `server` is subscribed to. 

`Channel.close` is also removed: `PeerGroup.client` returns a `Resource` that releases itself if left to go out of scope (whether that release was just book keeping of a shared resource or closing the channel is up to the PeerGroup implementation). On the incoming server events the release was added to `ChannelCreated`. 

I noticed that in the downstream project the channel is being passed around a lot for the handshake and to the actors, with closing happening at arbitrary places. I think there I'll keep wrapping the channel in a specialized class that retains the `close` method. (On a side note: I thought maybe the handshake should be part of a PeerGroup interface itself, and `server` should yield already shaken channels, with the same going for `client`).

I know this was an optional task but I couldn't think of writing the `StaticUDPPeerGroup` without this kind of pattern.


scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -48,14 +45,6 @@ trait Channel[A, M] {
...
     */
   def in: ConnectableObservable[ChannelEvent[M]]
-
-  /**
-    * Terminate the Channel and clean up any resources.
-    *
-    * @return Implementations should wait for termination of the underlying network resources
-    *         and return a successful Task.
-    */
-  def close(): Task[Unit]
```
KonradStaniec, line 58: I am not sure about this. In my opinion possiblity to close a channel at arbitrary moment is quite fundamental and should be avialable without a lot of fuss. I am aware that we can do `.allocated` on resource but it seems quite convoluted. Even `fs2` defines close method on their `Socket` trait.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -144,10 +119,12 @@ object PeerGroup {
       * only to A or to the whole group that is referenced by the multicast address. Different
       * implementations could handle this logic differently.
       */
-    case class ChannelCreated[A, M](channel: Channel[A, M]) extends ServerEvent[A, M]
+    case class ChannelCreated[A, M](channel: Channel[A, M], release: Release) extends ServerEvent[A, M]
```
KonradStaniec, line 122: we would not need that if we would leave `close` in channel interface

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -119,31 +142,51 @@ class ReqResponseProtocol[A, M](
...
+    Resource
+      .make(
+        for {
+          channelSemaphore <- Semaphore[Task](1)
+          channelMapRef <- Ref.of[Task, ChannelMap[A, M]](Map.empty)
+          fiberMapRef <- Ref.of[Task, Map[ChannelId, Fiber[Task, Unit]]](Map.empty)
+          protocol = new ReqResponseProtocol[A, M](group, channelSemaphore, channelMapRef, fiberMapRef)
+        } yield (protocol, channelMapRef, fiberMapRef)
+      ) {
+        case (_, channelMapRef, fiberMapRef) =>
```
KonradStaniec: this whole bit below is a little bit unreadable, maybe we can split it into named functions ?

KonradStaniec (commented):
In general great work!

two remarks:
1. This change will force a litte bit of unplanned work in downstream project when we decide to upgrade scalanet version, but if you say it was necessary for the next task then it is what it is I gueess.
2. 

> On a side note: I thought maybe the handshake should be part of a PeerGroup interface itself, and server should yield already shaken channels, with the same going for client

That was my general plan for this feature :) Although it would be rather specialised interface with additional type parameter, to account for additional peer info exchanged during handshake. If you check our downstream project you will see that I had defined channel which we receive from handshake as:
```
trait ChannelWithInfo[A, M, I] extends Channel[A, M] {
  def getAdditionalInfo: I
}
```

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: Yes this is why we need this here. I know in the downstream project it would be less convenient to pass it around so we can keep wrapping it as it is now, but I think as you'll see in the next PR it's actually much easier this way to attach extra logic that needs to happen when a channel is closed.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: I know it's a bit strange, but let's see how it works, we can always add it back. I found it more natural to do transformations and attach extra book-keeping that needs to happen when the channel is returned from the peer group. And it's much easier not to forget that to you have to call it because you don't have to remember it at all, or you have an extra variable that you have to think about how to use. Another benefit is that unless you go the advanced `.allocated` way there's just no way you'd call `sendMessage` on a closed channel, so overall finding out where a channel is closed becomes much easier.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
aakoshh: Okay, I'll move it into the class, thanks.

aakoshh:
> This change will force a litte bit of unplanned work in downstream project when we decide to upgrade scalanet version

Yeah, but not that much as I already have a huge PR to introduce resources there too. Plus it already wraps the class as you mentioned so it can keep adapting it to its needs, so I don't expect this will be a huge problem.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
aakoshh: Done :+1:

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
KonradStaniec: The thing is that restricting access to such fundamental operation on the upstream level, can lead to convoluted logic downstream just to get around those restrictions and just get the job done.  I know that, I would be highly unhappy if library forced me to go with this `.allocated` dance just to close channel. 

If you insist on it lets leave it as it is, but imo it just  passing the problem into the future i.e there will quickly be a task to add it back.

KonradStaniec (approved):

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: Thanks, I hear you. I think it all depends on the use case, the way the downstream project is treating the connection as something that changes hands all the time and eventually gets closed at an unknown location doesn't fit this pattern very much. But I thought this library's headline says it's "resource managed", so it should take the conservative approach.

PR #91, opened 2020-10-02
ETCM-162: StaticUDPPeerGroup

aakoshh:
Adds a `StaticUDPPeerGroup` that reuses the same local port for channels so that it can be compatible with Ethereum nodes.

I used cats effect concurrency primitives rather than concurrent collections, but in theory they could be implemented with the standard data structures, or even better with Monix Atomics, to avoid having to do the `runSyncUnsafe` in the channel handler; although I ended up using `runAsyncAndForget` since with UDP the ordering of messages doesn't matter. I figured if we want to use a `monix.catnap.ConcurrentQueue` then it will be inevitable to have effects, and the creation of the classes is effectful as well because of the concurrency primitives they use inside, such as Semaphore (which are good as they are non-blocking). 

Anyway, this is one way, I suggest that if we don't like it then we open another ticket to do it in another way and perhaps compare benchmarks. For now I think we can move on to implementing Kademlia.

I'm sorry about the syntactic noise I had to add after enabling warnings about discarded values. This was necessary because the compiler didn't complain about a case like `def handle(data: DatagramPacket): Unit = Task(whatever(data))` which would have left the task lost. At least it makes it clear where we're kicking off futures as well. But if you think it's not worth it, we can rely on tests and code reviews, I only wanted this to catch anything I forget in the Netty handlers. Or we can add an `ignore` function like F#.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
@@ -204,8 +204,9 @@ class StaticUDPPeerGroup[M] private (
   // The event loop will wait, but we have to delegate the execution elsewhere.
   // The alternative would be to use the normal concurrent data structures
   // that do internal locking without Tasks.
-  private def executeSync(task: Task[Unit]) =
+  private def executeSync(task: Task[Unit]): Unit = {
     task.runSyncUnsafe()
+  }
 
   private def tryDecodeDatagram(datagram: DatagramPacket): Attempt[M] =
     codec.decodeValue(BitVector(datagram.content.nioBuffer)) match {
@@ -269,11 +270,14 @@ class StaticUDPPeerGroup[M] private (
 
   // Wait until the server is bound.
   private def initialize(): Task[Unit] =
-    raiseIfShutdown >>
-      toTask(serverBinding).onErrorRecoverWith {
+    for {
+      _ <- raiseIfShutdown
+      _ <- toTask(serverBinding).onErrorRecoverWith {
         case NonFatal(ex) =>
           Task.raiseError(InitializationError(ex.getMessage, ex.getCause))
-      } >> Task(logger.info(s"Server bound to address ${config.bindAddress}"))
+      }
+      _ <- Task(logger.info(s"Server bound to address ${config.bindAddress}"))
+    } yield ()
 
   private def shutdown(): Task[Unit] = {
     for {
@@ -395,7 +399,7 @@ object StaticUDPPeerGroup extends StrictLogging {
         for {
           _ <- raiseIfClosed
           _ <- isClosedRef.set(true)
-          _ = messageSubject.onComplete()
+          _ <- Task(messageSubject.onComplete())
         } yield ()
       }
 
@@ -406,7 +410,12 @@ object StaticUDPPeerGroup extends StrictLogging {
   }
 
   private object ChannelImpl {
-    sealed trait Role
+    sealed trait Role {
+      override def toString(): String = this match {
+        case Server => "server"
+        case Client => "client"
+      }
+    }
     object Server extends Role
     object Client extends Role
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
@@ -59,6 +59,7 @@ class StaticUDPPeerGroup[M] private (
     workerGroup: NioEventLoopGroup,
     isShutdownRef: Ref[Task, Boolean],
     serverSubject: ConnectableSubject[ServerEvent[InetMultiAddress, M]],
+    serverChannelSemaphore: Semaphore[Task],
     serverChannelsRef: Ref[Task, Map[InetSocketAddress, StaticUDPPeerGroup.ChannelAlloc[M]]],
     clientChannelsRef: Ref[Task, Map[InetSocketAddress, Set[StaticUDPPeerGroup.ChannelAlloc[M]]]]
 )(implicit scheduler: Scheduler, codec: Codec[M])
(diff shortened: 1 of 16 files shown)


scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
```diff
@@ -0,0 +1,444 @@
...
+                  datagram.content().release()
+                  ()
+                }
+              }
+
+              override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
+                val channelId = ctx.channel().id()
+                val remoteAddress = ctx.channel.remoteAddress().asInstanceOf[InetSocketAddress]
+                cause match {
+                  case _: PortUnreachableException =>
```
KonradStaniec: this case is probably not necessary as it was thrown only when using client bootstrap and doing connected version of udp.

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
```diff
@@ -0,0 +1,444 @@
...
+      toTask(group.shutdownGracefully())
+    }
+
+  private class ChannelImpl[M](
+      nettyChannel: io.netty.channel.Channel,
+      localAddress: InetSocketAddress,
+      remoteAddress: InetSocketAddress,
+      messageSubject: ConnectableSubject[ChannelEvent[M]],
+      // Prevent race conditions between closing the channel and publishing messages.
+      subjectSemaphore: Semaphore[Task],
```
KonradStaniec, line 344: I wonder if we could get rid of this semaphore by using closing scheme similar to `DynamicUDPPeerGroup` i.e 
closing only  completes netty promise, and listener (which is also working on netty event loop) removes channel from map. Then there is no racing invloved as the thread which intercept the message, and closes channel is the same thread.

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
```diff
@@ -0,0 +1,444 @@
...
+  * request type message for a long time.
+  *
+  * @tparam M the message type.
+  */
+class StaticUDPPeerGroup[M] private (
+    config: StaticUDPPeerGroup.Config,
+    workerGroup: NioEventLoopGroup,
+    isShutdownRef: Ref[Task, Boolean],
+    serverSubject: ConnectableSubject[ServerEvent[InetMultiAddress, M]],
+    serverChannelsRef: Ref[Task, Map[InetSocketAddress, StaticUDPPeerGroup.ChannelAlloc[M]]],
```
KonradStaniec, line 63: why do we need separate refs for server channels and client channels ? If lets say peer1 opens client channer to peer2 , and peer2 sends message to peer1, shoudn't this message lands on this client channel and create new server channel ?

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
aakoshh: Okay great, I'll remove it, thanks!

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
aakoshh: I see what you mean you'd want the `onCompleted` to be called from somewhere else. I don't really want to complicate it for two reasons: the semaphore isn't blocking so it's probably not slowing anything down, and the next PR removes the `ConnectableSubject` anyway so we won't have this problem. It's also likely that the Observables protect against this too, so this is just being overly cautions, but adding netty promises to the mix sounds complicated.

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
aakoshh: That sounds correct, if peer1 opens a client to peer2 and sends a message, then peer2 will receive it on a new server channel. If peer1 sends more messages they arrive on this channel without a new server channel being created. So to handle incoming requests all you have to do is spawn a handler fiber for each new server channel. The responses peer2 sends on that server channel will go to peer1, and here it's going to be a bit unintuitive: peer1 will see the message on a new server channel as well as the client channel it has open. This is because there's generally no way to say why peer2 is sending a message.

If now peer2 opens a client to peer1, it will have one or more client channels too. They will all have the messages replicated, so it can use that client to just collect the messages that look like responses to what it sent on that client.

We need to keep these separate because for an incoming message there must be at most one server channel to handle all requests, so it can do things like accumulate state. If there's already an outgoing client channel connected to the same address, it may not expect the message coming in right now, so while we have to send it there, it's unlikely to handle it properly. 

Something like this:

```scala
peer1.server.foreachL {
  case ChannelCreated(channel, release) =>
    channel.in.collect {
       case MessageReceived(m: Request) => 
          channel.sendMessage(response(m))
    }
   .guarantee(release)
   .startAndForget
}

peer1.client(peer2.processAddress).use {
  client.sendMessage(request) >> client.in.collect {
     case MessageReceived(m: Response) if m.requestId == request.requestId =>
       m
  }.headL.timeout(5.seconds)
}
```

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
aakoshh: Removed, still calling `super`.

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
KonradStaniec: hmm that when i do not agree as:
- using netty concurrency primitives in not complicating, as rather natural way to interface with netty. I like the monix approach to this things i.e dirty inside and clean interface outside.
- any maybe overhead of semaphore is not much, but it is on lowest level of the stack. 
I do not want to block rest of the work, but imo after removing `ConnectableSubject` we should maybe revisit is.

aakoshh: I agree that there's room to optimise and I appreciate your suggestions a lot, it would just probably take me too long to figure it out now. I know we can do a completely different implementation of this whole thing and now there are a few quick and dirty `runSyncUnsafe` instances which I hate. I do find the cats-effect machinery easier to reason about though, and I hope that on the long run we can find better integration patterns. FWIW the semaphore is removed in the next PR, and I changed this one to do the message handling asynchronously.

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
KonradStaniec: I see. I am not sure about this but this whole abstraction it seems complex for general purpose udp thing as:
- it is pretty unintuitive
- it also makes some assumptions about how it will be used (request/response)
I would say that that we need next set of eyes on this, and maybe it is time to consider simple wrapper over netty udp like:
```
trait Udp {
 send(m): Tasl[Unit]
 incomining: Observble[M]
}
```
and go from there, as i fear we are just trying to fit square peg in round hole forcoing udp to fit in this channel abstraction.

aakoshh: Yes this is the only way to make it work with this abstraction and the kind of protocol Ethereum is using, i.e. there's no request-id in responses. That simple trait you are proposing is how Netty works, however if we used that, then the same kind of channel-replication machinery would have to be done in the application level, for example the the [ExpectedResponseChannels](https://github.com/ethereum/trinity/blob/dc0ddbb5ee5129b29ea2d4a9444f6a1b60cf09f4/p2p/discovery.py#L1352) in `trinity` and the [replyMatcher](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/v4_udp.go#L94) in `go-ethereum`.

The current approach appeals to me because even though it might be unintuitive, the actual code that is required to handle stuff looks exactly the same as we do for the other implementation, e.g. [KNetwork.sendRequest](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/kademlia/src/io/iohk/scalanet/kademlia/KNetwork.scala#L100). So you don't have to be aware, as long as you have a reasonable protocol where you can distinguish request and responses.

aakoshh: It wouldn't be unreasonable to think that at some point we might want to add options to make this class more knowledgeable and able to pair up channels, but the making it more efficient, while keeping the interface. But if we give up on the `PeerGroup` abstraction I'm not sure what reason we have left to bother with Scalanet in this case.

aakoshh: On the positive side https://github.com/input-output-hk/scalanet/pull/93 worked with almost no changes.

KonradStaniec (approved):
It seems no one has time to add another opinion on this, so lets go that way and see how it goes.

PR #92, opened 2020-10-05
ETCM-170: Use a message queue to publish events

aakoshh:
Buids on https://github.com/input-output-hk/scalanet/pull/91 see the [diff](https://github.com/input-output-hk/scalanet/compare/ETCM-162-static-udp-peergroup...ETCM-170-message-queue). 

Based on the discussions in https://github.com/input-output-hk/scalanet/pull/80 the PR adds a `CloseableQueue` which we can poll using `Channel.nextMessage()`. This should simplify some use cases and make it easier to reason about the state of the channel. In particular we don't have to think as hard about whether `channel.in.connect()` has been called yet or whether we accidentally dropped the last subscriber from `channel.in.refCount`, severing the connection. The PR doesn't rewrite all the existing code, the `Channel` still has an `in` property that works as it did so far, so the feature is opt-in at this point. It can make things like handshakes easier to implement, without having to worry about the setup sequence before connecting and the internal buffering of Observables.

I set a default channel capacity to unlimited but plan to use some value with UDP - it's connectionless and there are no guarantees anyway. 

We could opt to raise server events asynchronously, but currently it's all blocking.


aakoshh:
@jtownson this should make it less fiddly to consume events, no need to `connect` any more. In the TCP implementation we plan to track the remaining capacity of the queues and set the Netty socket to non-auto-read mode if it reaches the limits, stopping the incoming data until messages are consumed.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -41,10 +43,22 @@ trait Channel[A, M] {
...
+  /**
+    * Consume the next message from the underlying event queue.
+    *
+    * @return The next incoming message, or None if the channel was closed.
+    */
+  def nextMessage(): Task[Option[ChannelEvent[M]]]
+
   /**
     * The inbound stream of messages coming from the remote peer.
+    * Provided for backwards compatibility.
```
KonradStaniec, line 55: I wonder do we really need it ? we already have changes interfaces enugh that next update on downstream project will be a lot of work, so forcing us to manaully do `ConnectableSubject.repeatEvalNext(nextMessage())(s)` is not big deal. And this forces this ugly scheduler at interface level.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -75,8 +76,14 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
...
       )
-      messageSubject.onNext(UnexpectedError(cause))
+      handleEvent(UnexpectedError(cause))
     }
+
+    private def handleEvent(event: ChannelEvent[M]): Unit =
+      executeSync(messageQueue.tryOffer(event).void)
+
+    private def executeSync(task: Task[Unit]): Unit =
+      task.runSyncUnsafe()
```
KonradStaniec: I wonder, shouldn't we push it on netty event loop of this channel to satisfy single producer scenario ? As i understand when we do push it on this provided scheduler diffrent threads can do the job, there will be always at most one producer, but each time it could be different thread.

scalanet/src/io/iohk/scalanet/peergroup/CloseableQueue.scala
```diff
@@ -0,0 +1,95 @@
...
+          case Right(item) =>
+            Task.pure(Some(item))
+        }
+    }
+
+  /** Stop accepting items in the queue. Clear items if `discard` is true, otherwise let them be drained.
+    * If the queue is already closed it does nothing; this is because either the producer or the consumer
+    * could have closed the queue before.
+    */
+  def close(discard: Boolean): Task[Unit] =
```
KonradStaniec, line 50: i find booleans in public interfaces as something really confusing. Could't we provide two methods for this.?

scalanet/src/io/iohk/scalanet/peergroup/CloseableQueue.scala
```diff
@@ -0,0 +1,95 @@
...
+    closed.complete(discard).attempt >> queue.clear.whenA(discard)
+
+  /** Try to put a new item in the queue, unless the capactiy has been reached or the queue has been closed. */
+  def tryOffer(item: A): Task[Either[Closed, Boolean]] =
```
KonradStaniec, line 60: maybe it is matter of opinion, why not returning failed task when some one tries to put item in closed queue ?

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: I just didn't want to go and modify all the tests right now, it was good to show that everything still works. We can totally add a new ticket to remove it altogether.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: I'm glad you asked, I hoped to get your input! I tried your [technique](https://github.com/input-output-hk/scalanet/pull/80/files#diff-a967c65483f7c8c1745cf0cbdcc22900R198-R201) first which is to publish on `val eventLoopAsScheduler = monix.execution.Scheduler(nettyChannel.eventLoop())` with `task.runSyncStep(eventLoopAsScheduler)`, however the tests deadlocked, and I thought there were two problems with this:

1. First is that [Task.runSyncStep](https://monix.io/api/3.0/monix/eval/Task.html#runSyncStep(implicits:monix.execution.Scheduler):Either[monix.eval.Task[A],A]) only executes up to the first async boundary, and it returns either the computed result or a task which represents the async calculation. In this case when we just do the `tryOffer` it may fit in one step, I don't know, but the whole stuff that happens in `StaticUDPPeerGroup` didn't. 
2. When I tried `task.runSyncUnsafe(eventLoopAsScheduler, implicitly[CanBlock])` it still deadlocked. I figured the reason was that the `channel.eventLoop` is basically one thread which is currently executing what it does now, which is handling the Netty event. We can't ask it to schedule in our task as well and wait on it, because it gets in the queue of this very thread, it doesn't execute on it straight away, unlike `runSyncStep`. I tried various combinations like executing the data structure maintenance synchronously on the netty scheduler and the replication on the implicit one.

We could definitely execute on the `workGroup` as that one has multiple threads since it's an `EventLoopGroup`, not an `EventLoop`. 

But that would still not answer your question about whether `SPMC` is appropriate and the answer is that I don't really know, I haven't checked under the hood. I don't know if it means that you can only publish from one thread, or whether you cannot publish concurrently.

scalanet/src/io/iohk/scalanet/peergroup/CloseableQueue.scala
aakoshh: Sure, I'll add methods for it. I always use named arguments for this kind of stuff but I see your point.

scalanet/src/io/iohk/scalanet/peergroup/CloseableQueue.scala
aakoshh: Because it's not an exceptional condition, it's better in this case on a "closeable queue" to force people to think about the possibility that the offering will not be taken. I wouldn't want the producer to bomb just because the consumer isn't interested in the data any more and they forgot to catch the exception: they should check why the offering failed (closed or throttled) or ignore it because there's nothing they can do about it anyway (the channel itself is being closed in our case so the producer is going to go away too). Similarly the underlying `tryOffer` doesn't throw an exception, it returns a boolean that you may or may not check.

My rule of thumb is that if the method is expected to fail, such as a `try`, or a validation function, then it's good to use an `Either` to represent the types (e.g. I wouldn't know what exception to look for just by looking at the type), and I make the error types inherit `NoStackTrace` so the caller can easily call `.rethrow` on the thing if they rather want just exceptions than handle it the cumbersome way. In this case I didn't do that because I didn't think it was an error to try to push into a closed queue, it was normal program flow.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Hm, just read that `queue.clear` should only be called by the consumer, will fix that.

aakoshh: Looks like [MpmcArrayQueue.offer](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java#L194-L215) executes a loop while [SpmcArrayQueue.offer](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpmcArrayQueue.java#L206-L233) doesn't.

So as long as the checks themselves do a memory fence they should get up-to-date indexes, and since there's no other producer manipulating the queue there shouldn't be a need for a loop.

KonradStaniec: hmm which tests deadlocked ? (I will try to play with it locally a bit)

I am also thinking, what would be the result of just callit `messageQueue.tryOffer(event).runToFuture` (or other async runner) on channel eventloop. In theory when executing tasks on single threaded scheduler they are guaranteed to execute in order of scheduling as underneath single thread scheduler is using linked blocking queue to execute tasks in order. So messages should be delivered in order.
Also on udp peer groups we could probably do that either way as udp in principle does not care about ordering

scalanet/src/io/iohk/scalanet/peergroup/CloseableQueue.scala
KonradStaniec: oh that makes sense. i.e the whole point of having `closeable queue` is to make closing representable errror condition.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Yeah the `StaticUDPPeerGroup` now starts things in the background because the order doesn't matter, as you said. Let me run it, I can't remember if the deadlock is still present.

aakoshh: For example with this change the tests time out:

```scala
  private def executeAsync(task: Task[Unit]): Unit = {
    //task.runAsyncAndForget
    import monix.execution.schedulers.CanBlock
    task.runSyncUnsafe()(Scheduler(serverBinding.channel().eventLoop()), implicitly[CanBlock])
  }
```

```console
$ mill scalanet.test.single io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroupSpec
[54/55] scalanet.test.runMain 
Run starting. Expected test count is: 15
StaticUDPPeerGroupSpec:
StaticUDPPeerGroup
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup INFO  Server bound to address localhost/127.0.0.1:46237
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup DEBUG Added UDP client channel from localhost/127.0.0.1:46237 to localhost/127.0.0.1:43195
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup$ChannelImpl DEBUG Sending client message ByteVector(16777216 bytes, #-648541556)... from localhost/127.0.0.1:46237 to localhost/127.0.0.1:43195
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup DEBUG Removed UDP client channel from localhost/127.0.0.1:46237 to localhost/127.0.0.1:43195
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup INFO  Shutting down UDP peer group for peer InetMultiAddress(localhost/127.0.0.1:46237)
- should report an error for sending a message greater than the MTU
test-17 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup INFO  Server bound to address localhost/127.0.0.1:34353
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup INFO  Server bound to address localhost/127.0.0.1:38099
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup DEBUG Added UDP client channel from localhost/127.0.0.1:34353 to localhost/127.0.0.1:38099
test-14 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup$ChannelImpl DEBUG Sending client message Hi Bob!... from localhost/127.0.0.1:34353 to localhost/127.0.0.1:38099
nioEventLoopGroup-4-1 io.iohk.scalanet.peergroup.udp.StaticUDPPeerGroup DEBUG Server channel at localhost/127.0.0.1:38099 read message from /127.0.0.1:34353
- should send and receive a message *** FAILED ***
  java.util.concurrent.TimeoutException: Task.runSyncUnsafe(15 seconds)

```

aakoshh: What you suggested seems to work:
```scala
  private def executeAsync(task: Task[Unit]): Unit = {
    task.runAsyncAndForget(Scheduler(serverBinding.channel().eventLoop()))
  }
```

aakoshh: Good idea, shall we do this on the TCP groups then?

jtownson (approved):
This PR (and the earlier PoC) brings a smile to my face. It's definitely a step closer to having an intuitive and easy to use networking library. Look forward to the `ConnectableSubject` instance being removed in future.

I don't know if scalanet has any external users, but it is probably worth putting @deprecated annotations on that stuff.

aakoshh:
@jtownson great to hear! 

> I don't know if scalanet has any external users, but it is probably worth putting @deprecated annotations on that stuff.

I thought about that but I'd have had to rewrite everything that uses it and the tests (can't even compile with the warnings in with the current settings). Konrad thinks we can just remove it so I'll make a ticket to go over it once I'm finished with the discovery story.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: hmm we would have to be sure that messages are provided in correct order, maybe we could first add some test for it ? (also we will probably go through this again when adding back pressoure but this  is future worry)

aakoshh: Okay done, works dandy, great stuff :+1:

KonradStaniec (approved):

PR #94, opened 2020-10-07
ETCM-164: Restrict scodec implicits

aakoshh:
Making sure that `DefaultCodecs` can provide all the necessary `scodec.Codec` instances without the user having to import `scodec.codecs.implicits._`, which would bring in the opinionated codecs that wouldn't be compatible with Ethereum. The import from `DefaultCodecs` is done outside the components we want to use, so for Ethereum we should be able to provide RLP specific `Codec` instances instead of what's in `DefaultCodecs`. Not having to import any more implicits should make it obvious which one is used.


aakoshh:
I don't know why the scalafmt installation step fails on CircleCI.

KonradStaniec (approved):

aakoshh:
Merged as part of https://github.com/input-output-hk/scalanet/pull/95/commits

PR #95, opened 2020-10-09
ETCM-196: Discovery v4 Part 1

aakoshh:
Builds on https://github.com/input-output-hk/scalanet/pull/93 see the [diff](https://github.com/input-output-hk/scalanet/compare/ETCM-163-kademlia-use-static-udp...ETCM-196-discovery-v4-part1)

Part 1 of the Discovery story aims to add the logic to implement the spec without RLP or proper cryptography, those will be mocked or substituted with scodec. The task is still big, so this is the first PR adds just a `DiscoveryNetwork` component, some crypto traits it needed to function and the protocol messages (`Payload` and `Packet`). The `DiscoveryNetwork` is similar to the `KNetwork` in that it encapsulates logic of sending and receiving messages from the network. 

I started off with a similar pattern as the `KNetwork` and returned a stream of requests which gets processed in parallel by whoever consumes them. I wanted to make it a bit more type safe so that callbacks only accept the right kind of response, and I don't have to pass the whole `Packet` but just the bits that are relevant, so it wouldn't be `Message => Message` but the specific pairs like `PublicKey => Seq[Node]`. It worked but it was a bit complicated, and there were some issues I didn't like about the streaming approach:
* There's no guarantee that the callback won't be called multiple times, which wouldn't make any sense.
* There's no guarantee that the channel is still open at the time the callback is executed.
* If the consumer holds on to the callback it prevents the channel from being garbage collected.
* It didn't work with an `Iterant` because that lacks the kind of semantics `Observable.mergeMap` has. 
* The stream of requests executed concurrently could be dominated by one overproducing peer starving out everybody else.
* There was a buffering strategy on top of the common stream dropping messages over the limit, which again could be an attack vector by an overproducing peer. It also defeats the buffer limits of the underlying queues.

I inverted the pattern and pass in a handler to the network that handles the same RPC calls as the network itself: `ping`, `findNode`, and `enrRequest`. The network starts a fiber to handle incoming requests and further fibers to handle each individual channel. This has a couple of advantages:
* It's type safe, we know it won't mix up request and response pair types.
* It's fair: each remote peer gets their requests handled one by one, while peers are handled concurrently. 
* Allows using request type specific semaphores for throttling for example `findNode` across all peers.
* Because the network component is calling the handler, we know it will only call it exactly once.
* There's no common buffer overflow, it's handled on the per-channel queue level, so an overproducing peer doesn't cause degradation for the others.


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+            .map(_.map(_._1))
+
+      }
+    }
+  }
+
+  // Functions to be applied on the `.nextMessage()` or `.nextServerEvent()` results.
+  private implicit class NextOps[A](next: Task[Option[A]]) {
+    def toIterant: Iterant[Task, A] =
+      Iterant.repeatEvalF(next).takeWhile(_.isDefined).map(_.get)
```
KonradStaniec, line 365: Why `Iterant` ? I mean i am not familiar with it (only aware that it is more in spirit of fs2 pull based stream), but we already have observables, subjects etc. What additional things iterant gives us ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+          .toIterant
+          .mapEval {
+            case MessageReceived(packet) =>
+              currentTimeMillis.flatMap { timestamp =>
+                Packet.unpack(packet) match {
+                  case Attempt.Successful((payload, remotePublicKey)) =>
+                    payload match {
+                      case _: Payload.Response =>
+                        // Not relevant on the server channel.
+                        Task.unit
```
KonradStaniec, line 116: shouldn't we return error or at leas log error, as either we have:
- unrequested response 
- somthing went bad in underlaying peergroup
both of this conditions are at least interesting imp

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+object DiscoveryNetwork {
+
+  // Errors that stop the processing of incoming messages on a channel.
+  class PacketException(message: String) extends Exception(message) with NoStackTrace
+
+  def apply[A](
+      peerGroup: PeerGroup[A, Packet],
+      privateKey: PrivateKey,
+      toNodeAddress: A => Node.Address,
+      messageExpiration: FiniteDuration = 60.seconds,
```
KonradStaniec: maybe we could wrap up following values in some kinda `Config` case class ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+            case MessageReceived(packet) =>
+              currentTimeMillis.flatMap { timestamp =>
+                Packet.unpack(packet) match {
+                  case Attempt.Successful((payload, remotePublicKey)) =>
+                    payload match {
+                      case _: Payload.Response =>
+                        // Not relevant on the server channel.
+                        Task.unit
+
+                      case p: Payload.HasExpiration[_] if p.expiration < timestamp - expirationMillis =>
```
KonradStaniec: I remember that geth had many problem with not working discovery due to badly configured system clock on user computer (and many timeouted messages due to that)
Maybe w can create task to contact use npt to check if user time is correctly configured warn him about this problem ?
And for now maybe we can have an option to disable expriation checks ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+      import DiscoveryRPC.ENRSeq
+      import Payload._
+
+      private val expirationMillis = messageExpiration.toMillis
+
+      private val currentTimeMillis = clock.realTime(MILLISECONDS)
+
+      private val maxNeighborsPerPacket = getMaxNeighborsPerPacket
+
+      private val localNodeAddress = toNodeAddress(peerGroup.processAddress)
```
KonradStaniec, line 65: I am not sure if this should be taken out of `peerGroup.processAddress` as in most cases this returns actually interface with which the peergroup has been started not acutual external ip address i.e most probably `127.0.0.1` in test envs or `0.0.0.0` in real working enviroment. And in ping message we probably want to use real ip address (although i am not sure that important here, as we do not respond to advertised address but to the address we got packet form)

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+              currentTimeMillis.flatMap { timestamp =>
+                Packet.unpack(packet) match {
+                  case Attempt.Successful((payload, remotePublicKey)) =>
+                    payload match {
+                      case _: Payload.Request =>
+                        // Not relevant on the client channel.
+                        Task.pure(None)
+
+                      case p: Payload.HasExpiration[_] if p.expiration < timestamp - expirationMillis =>
+                        Task.pure(None)
```
KonradStaniec: Shouldn't we at least log expirations ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+      override val ping = (to: A) =>
+        (localEnrSeq: Option[ENRSeq]) =>
+          peerGroup.client(to).use { channel =>
+            channel
+              .send(
+                Ping(version = 4, from = localNodeAddress, to = toNodeAddress(to), 0, localEnrSeq)
+              )
+              .flatMap { packet =>
+                channel.collectFirstResponse {
+                  case Pong(_, pingHash, _, maybeRemoteEnrSeq) if pingHash == packet.hash =>
```
KonradStaniec, line 234: just of caution, did you make sure this wat geth has in pong response ? (and maybe besu:https://github.com/hyperledger/besu)

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,376 @@
...
+                }
+              }
+          }
+
+      override val findNode = (to: A) =>
+        (target: PublicKey) =>
+          peerGroup.client(to).use { channel =>
+            channel.send(FindNode(target, 0)).flatMap { _ =>
+              channel.collectAndFoldResponses(kademliaTimeout, Vector.empty[Node]) {
+                case Neighbors(nodes, _) => nodes
```
KonradStaniec, line 245: from what i see geth sanitizes received nodes based on the fact if they are received from local node or remote node: 
https://github.com/ethereum/go-ethereum/blob/723bd8c17fdf1614d81503efdca9e7eece3766a6/p2p/discover/v4_udp.go#L727.
Maybe we can create the task to do the same ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryRPC.scala
```diff
@@ -0,0 +1,58 @@
...
+object DiscoveryRPC {
+  type ENRSeq = Long
+
+  /** Pair up requests with responses in the RPC. */
+  sealed trait Proc {
+    type Req
+    type Res
+  }
+  object Proc {
+    trait Ping extends Proc {
```
KonradStaniec, line 36: I find it a little not intuitive that we alreagy have  already concrete `Ping` case class and we need to redifine it here as trait

scalanet/discovery/test/src/io/iohk/scalanet/discovery/crypto/SigAlgSpec.scala
```diff
@@ -0,0 +1,8 @@
+package io.iohk.scalanet.discovery.crypto
+
+import org.scalatest._
+
+class SigAlgSpec extends FlatSpec with Matchers {
+  // Use test vectors from https://wizardforcel.gitbooks.io/practical-cryptography-for-developers-book/content/digital-signatures/ecdsa-sign-verify-examples.html
+  // Implement recovery based on https://github.com/ConsenSysMesh/cava/blob/master/crypto/src/main/java/net/consensys/cava/crypto/SECP256K1.java
```
KonradStaniec, line 7: hm we already has recovery in mantis and whole (ECDSA key handling) is this somewhat better ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Iterant is entirely pull based, so there's no need for stuff like worrying about back pressure or buffering. Since we have a method that reads one message at a time, I think this is easier to reason about in this context, when we pull messages from a queue.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: No this isn't an error here, this is how the `StaticUDPPeerGroup` works. Similarly to how [KNetwork only collects KRequests](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/kademlia/src/io/iohk/scalanet/kademlia/KNetwork.scala#L58) here we are in the context of the server channel, so we can ignore messages which don't need answers. With the `DynamicUDPPeerGroup` we could have had an expectation that a client connecting to us will send requests, however in the case of the `StaticUDPPeerGroup`  we actually know that the server channel will receive _all_ incoming messages, not just requests, because it can never tell about an incoming packet what it is, they all arrive from the same socket, so it does the replication to all channels.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Yeah, something like that would be good, one on the whole `discovery.ethereum.v4` level so not ever class needs to get their own.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Indeed the spec says:
> The expiration field present in all packets is supposed to prevent packet replay. Since it is an absolute time stamp, the node's clock must be accurate to verify it correctly. Since the protocol's launch in 2016 we have received countless reports about connectivity issues related to the user's clock being wrong.

Some kind of check seems prudent and I guess you can always increase the expiration to ridiculous values like 24 hours - I doubt anyone can participate in blockchain if their computer's clock is off by that much. I took the default 1 minute from trinity. Funnily if we increase the expiration to higher the other side will accept it as well as long as its not in the past.

I saw some discussions about how bad it is to require a centralised authority like NPT in a decentralised application for correctness; it sounds like a good idea to warn about clock drift even if it's disabled. I think it should be in another component though, something that the node starts. It could be relevant to certain consensus algorithms as well.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: The concept of the `processAddress` is a bit nebulous to me. I saw that currently the `*PeerGroup.Config` classes basically use the `bindAddress` as the `processAddress`, but form the interface it looks like that its type `A` is actually addressable, otherwise how would `PeerGroup.client(to: A)` work? I thought actually `bindAddress` might be the one pointing at `0.0.0.0` and the `processAddress` would be the one you use to participate in the p2p network, the outwardly meaningful one. It seemed to make sense that if we use addresses of type `A` to connect to each other then these `A` addresses are what need to be sent to each other as well.

It will be important that this is correct in `Neighbor` packets, but you are right the ones in `Ping` are completely ignored. In the next PR I'm passing in a `Node` to the service, so we can do that here too if necessary.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Ah, I forgot that the the expiration check means if the other side is using a 1 minute expiration and we're off by more than a minute we can still ignore their messages, so increasing our timeout value doesn't help. Maybe there should be a setting to allow a maximum clock drift.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Yes, I'll add it.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: I didn't look at besu, just trinity. It had a special [workaround](https://github.com/ethereum/trinity/blob/dc0ddbb5ee5129b29ea2d4a9444f6a1b60cf09f4/p2p/discovery.py#L1028) for some openethereum 1.10 clients, but I figured since we'll use this in a new testnet in a fork, it won't have to work with old unupgraded clients. geth seems to [calculate](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/v4wire/v4wire.go#L267) the hash the same way and then [compare](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/v4_udp.go#L238) to the Pong.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Yes, thanks, I'll make a ticket.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryRPC.scala
aakoshh: I didn't want to come up with a new name, I only use it as `Proc.Ping` to distinguish, I thought it may be better than `PingProc` but I might be wrong. I wanted to capture the symmetric relationship between the remote and local RPC interfaces and the `Call` semantics without having to repeat the call signatures everywhere.

scalanet/discovery/test/src/io/iohk/scalanet/discovery/crypto/SigAlgSpec.scala
aakoshh: Yes I realised later that it must be available in Mantis, just left the comment as a reminder in case it's not. What I was planning to do is provide the implementation at the time when we instantiate this dependency within Mantis. Talked to @mirkoAlic yesterday about extracting RLP and Crypto; he's of the opinion that it should be polished a bit before we do it and it would be 1-2 weeks for just RLP.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Added a `maxClockDrift` setting, that can basically disable checks.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Added a `DiscoveryConfig`.

scalanet/discovery/test/src/io/iohk/scalanet/discovery/crypto/SigAlgSpec.scala
aakoshh: At the time I was looking at whether [CryptoUtils](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala) could be used to implement the `SigAlg`.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: It also has a killer [foldWhileLeftEvalL](https://monix.io/api/3.0/monix/tail/Iterant.html#foldWhileLeftEvalL[S](seed:F[S])(op:(S,A)=%3EF[Either[S,S]])(implicitF:cats.effect.Sync[F]):F[S]) method, although I didn't need it here.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
KonradStaniec: yep maybe we can think about it next. For now we have exteranal address discovery implemented on node level, maybe it would be worthwhile to move it scalanet ? then peer groups would be able to privde bind address and real external address ? just something to think about.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
KonradStaniec: we will use this in new test net and on mainnet. So i would check out at least compatibility with geth and besu.(as they make about 90% of etc nodes) openethereum announecd that it removes support for etc so it is probably not that important.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -0,0 +1,398 @@
...
+            .mapEval {
+              case ChannelCreated(channel, release) =>
+                handleChannel(handler, channel, cancelToken)
+                  .guarantee(release)
+                  .onErrorRecover {
+                    case ex: TimeoutException =>
+                    case NonFatal(ex) =>
+                      logger.error(s"Error handling channel from ${channel.to}: $ex")
+                  }
+                  .startAndForget
```
KonradStaniec, line 88: As we are starting concurrent fiber for each incoming channel, shouldn't we use some semaphore to limit the concurrency ? Otherwise it is easy to use up server resources by just sending messages from different adressess.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Yes I think that component about opening firewall holes and discovering external addresses would be a great addition to Scalanet, it's not something I'd know how to do on my own, would be great if the library catered for it, great idea :+1:

KonradStaniec: I will create a task to migrate those thing from our other project. There we got external address discovery, and upnp to automatic port forwarding in case user is behind the nat.

aakoshh: I changed this in https://github.com/input-output-hk/scalanet/pull/96/commits/5e9e33f56cfad26381aceb1236c7c632568b9c57 to pass the `localNodeAddress` directly; `besu` seems to be [picking](https://github.com/hyperledger/besu/blob/master/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java#L190) the TCP port from it.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: I checked besu, the [computation](https://github.com/hyperledger/besu/blob/f9deb54682cb24779bc4919f00b4a953763db8a9/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/Packet.java#L59-L60) of the packet hash is mostly the same, except for example that they hash the data with keccak256 before signing, where as I imagine [trinity](https://github.com/ethereum/trinity/blob/dc0ddbb5ee5129b29ea2d4a9444f6a1b60cf09f4/p2p/discovery.py#L1323-L1324) takes the hash inside the signing function itself, which is what our `SigAlg.sign` will have to do as well. 

It will also have to make sure that the `Signature` it returns is encoded in the expected way, compatible with [encodeSignature](https://github.com/ethereum/trinity/blob/dc0ddbb5ee5129b29ea2d4a9444f6a1b60cf09f4/p2p/discovery.py#L1323-L1324). I want to use the test vectors I found here and there but at the moment I can only do this in Mantis.

After that the [checking of the pong response](https://github.com/hyperledger/besu/blob/master/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java#L462) is the same.

aakoshh: On a side note: besu doesn't seem to [respect](https://github.com/hyperledger/besu/blob/master/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java#L555-L556) the packet size limits so we may never accept a neighbors response from them.

I see now that Trinity doesn't actually enforce the max packet size, but rather uses a 2*1280 byte buffer for reading UDP packets. We could do something similar with [FixedRecvByteBufAllocator](https://netty.io/4.0/api/io/netty/channel/FixedRecvByteBufAllocator.html) instead of [DefaultMaxBytesRecvByteBufAllocator](https://netty.io/4.1/api/io/netty/channel/DefaultMaxBytesRecvByteBufAllocator.html) (I don't fully understand what it does TBH).

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Great question! I wanted to look up how much resource a fiber actually takes; it looks like it's basically a `Task` scheduled to run asynchronously with an attached Promise and a cancel token. I don't know how many tasks can execute concurrently. 

But we have to do something with the incoming message here that isn't blocking the `mapEval` otherwise we'd block processing all incoming connections. I guess we could track something like the number of connections from the same IP address and immediately release if there are more than some threshold. 

The fiber itself it will be closed down after a minute of inactivity.

I thought a semaphore could be used on the higher level component to for example limit the number of certain concurrent operations, which is another kind of protection against resource exhaustion, but it's not obvious how to make it fair.

In the case of discovery, to do anything heavy like FindNode first these addresses actually have to respond to a Ping. I guess it's not a bad waste of resources to ask a node to keep pinging random ports on your machine, so if we limited the number of connections from any given IP we could at least keep that in check.

KonradStaniec: llimiting number of concurrent incoming channels from any given ip could be probably done at peer group level, and we could try to use for it netty `AbstractRemoteAddressFilter`  (or maybe it modified version as orginal is designed for connection based transport like tcp, and not udp). 

I just wonder if we should have any limits in case of getting spammed from like 10000 or more diffent ip-s at once. Ultimatly fibers are run on underlying scheduler threads, so from user perspective this is one possibility to limit resources, but i wonder if we should't have option to limit resources usage at higher lever of abstraction like semaphore with 10k of permits (or number configurable by user) which is basically saying: that is at most concurrent connection i would like to have and block accpeting next connectionns until there will more free resources. (but again maybe it should be done at peer group level ?)

KonradStaniec: tldr: it pretty non trivial topic lets deal with it separatly, either in part 2 or in separate pr to properly handle resource usage acroess all discovery components (peergroup -> disovery-network -> discovery-service)  :)

KonradStaniec (approved):

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: I agree, let's open a new ticket for it. Thank you for the pointer to `AbstractRemoteAddressFilter`, looks like a good fit indeed! 

We can have semaphores to avoid being swamped by requests from 10,000 different legit addresses, all asking us to find neighbours all the time, it's easy to say we only respond to, say, 10 FindNode requests at any time, and move on if we can't get a permit within the request timeout. We can also add rate limits per individual channel, so they can only issue so many lookups per time interval.

PR #96, opened 2020-10-16
ETCM-207: Discovery v4 Part2

aakoshh:
Builds on https://github.com/input-output-hk/scalanet/pull/95 see the [diff](https://github.com/input-output-hk/scalanet/compare/ETCM-196-discovery-v4-part1...ETCM-207-discovery-v4-part2)

Adds a `DiscoveryService` component which hopefully implements the [spec](https://github.com/ethereum/devp2p/blob/master/discv4.md#recursive-lookup).

The main differences from `KRouter` are:
* Before we talk to a node we need to `bond` with it.
* Does not respond to nodes it hasn't bonded with, unless it's a ping, which triggers a bond.
* Doesn't touch the k-bucket timestamps in any other case as when a Pong is received. I don't see a problem if it did, I just followed what Trinity does for now.
* Nodes encountered during the recursive `lookup` are bonded with, otherwise they aren't followed or returned in the closest set.
* When looking up a node, we only return the result if the ENR could be fetched; just getting it from the lookup procedure itself is not enough (they could lie about the TCP port).
* When the rest of the system asks for nodes (for gossiping for example) they only get ones where we could fetch the ENR.
* It will consider contacting all `k` closest nodes, not just alpha.
* Supports changing the external address.
* Updates the node address if it talks from a new port.
* Validates signatures.
* If the bucket the peer goes into is full and the existing node cannot be evicted, it still inserts the new candidate into the node records, just not the routing table. This way it could be resolved in `getNode` without having to look it up again, although it won't be returned in `FindNode` requests. But it _will_ be returned by `getNodes` if the rest of the system needs candidates to gossip to.
* It doesn't return the bootstrap nodes for gossiping until their ENR can be resolved.

Main differences from Trinity:
* Doesn't have a persistent ENR store.
* Only performs alpha concurrent `FindNode`
* Allows multiple `lookup`s at the same time
* Maintains the bonded status based on ID, IP and UDP port, so a new address requires new bonding.

The class is all about manipulating internal state and doing async operations, the public interface exposes very little of this. The tests have full access to all the helper methods and the state, so that I can unit tests everything by setting up and verifying the state directly.

### DIFF
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
@@ -112,7 +112,9 @@ object DiscoveryService {
       // Last time a peer responded with a Pong to our Ping.
       lastPongTimestampMap: Map[Peer[A], Timestamp],
       // Deferred results so we can ensure there's only one concurrent Ping to a given peer.
-      bondingResultsMap: Map[Peer[A], BondingResults]
+      bondingResultsMap: Map[Peer[A], BondingResults],
+      // Deferred ENR fetches so we only do one at a time to a given peer.
+      fetchEnrMap: Map[Peer[A], Deferred[Task, Unit]]
   ) {
     def withLastPongTimestamp(peer: Peer[A], timestamp: Timestamp): State[A] =
       copy(lastPongTimestampMap = lastPongTimestampMap.updated(peer, timestamp))
@@ -120,8 +122,20 @@ object DiscoveryService {
     def withBondingResults(peer: Peer[A], results: BondingResults): State[A] =
       copy(bondingResultsMap = bondingResultsMap.updated(peer, results))
 
+    def withEnrAndAddress(peer: Peer[A], enr: EthereumNodeRecord, address: Node.Address): State[A] =
+      copy(
+        enrMap = enrMap.updated(peer.id, enr),
+        nodeMap = nodeMap.updated(peer.id, Node(peer.id, address))
+      )
+
     def clearBondingResults(peer: Peer[A]): State[A] =
       copy(bondingResultsMap = bondingResultsMap - peer)
+
+    def withEnrFetch(peer: Peer[A], result: Deferred[Task, Unit]): State[A] =
+      copy(fetchEnrMap = fetchEnrMap.updated(peer, result))
+
+    def clearEnrFetch(peer: Peer[A]): State[A] =
+      copy(fetchEnrMap = fetchEnrMap - peer)
 
     def removePeer(peer: Peer[A]): State[A] =
       copy(
@@ -144,7 +158,8 @@ object DiscoveryService {
       nodeMap = Map(node.id -> node),
       enrMap = Map(node.id -> enr),
       lastPongTimestampMap = Map.empty[Peer[A], Timestamp],
-      bondingResultsMap = Map.empty[Peer[A], BondingResults]
+      bondingResultsMap = Map.empty[Peer[A], BondingResults],
+      fetchEnrMap = Map.empty[Peer[A], Deferred[Task, Unit]]
     )
   }
 
@@ -346,24 +361,50 @@ object DiscoveryService {
       } yield ()
 
     /** Fetch a fresh ENR from the peer and store it. */
-    protected[v4] def fetchEnr(peer: Peer[A]): Task[Unit] =
-      rpc.enrRequest(peer)(()).flatMap {
-        case None =>
-          Task.unit
-
-        case Some(enr) =>
-          EthereumNodeRecord.validateSignature(enr, publicKey = peer.id) match {
-            case Attempt.Successful(true) =>
-              tryUpdateEnr(peer, enr)
-
-            case Attempt.Successful(false) =>
-              Task(logger.debug("Could not validate ENR signature!")) >>
-                removePeer(peer)
-
-            case Attempt.Failure(err) =>
-              Task(logger.error(s"Error validateing ENR: $err"))
+    protected[v4] def fetchEnr(peer: Peer[A]): Task[Unit] = {
+      val waitOrFetch =
+        for {
+          d <- Deferred[Task, Unit]
+          decision <- stateRef.modify { state =>
+            state.fetchEnrMap.get(peer) match {
+              case Some(d) =>
+                state -> Left(d)
+              case None =>
+                state.withEnrFetch(peer, d) -> Right(d)
+            }
           }
+        } yield decision
+
+      waitOrFetch.flatMap {
+        case Left(wait) =>
+          wait.get
+
+        case Right(fetch) =>
+          rpc
+            .enrRequest(peer)(())
+            .flatMap {
+              case None =>
+                Task.unit
+
+              case Some(enr) =>
+                EthereumNodeRecord.validateSignature(enr, publicKey = peer.id) match {
+                  case Attempt.Successful(true) =>
+                    tryUpdateEnr(peer, enr)
+
+                  case Attempt.Successful(false) =>
+                    Task(logger.debug("Could not validate ENR signature!")) >>
+                      removePeer(peer)
+
+                  case Attempt.Failure(err) =>
+                    Task(logger.error(s"Error validateing ENR: $err"))
+                }
+            }
+            .guarantee {
+              stateRef.update(_.clearEnrFetch(peer)) >>
+                fetch.complete(())
+            }
       }
+    }
 
     /** Try to extract the node address from the ENR record and update the node database,
       * otherwise if there's no address we can use remove the peer.
@@ -373,12 +414,7 @@ object DiscoveryService {
         case None =>
           Task(logger.debug(s"Could not extract node address from $enr")) >> removePeer(peer)
         case Some(address) =>
-          stateRef.update { state =>
-            state.copy(
-              enrMap = state.enrMap.updated(peer.id, enr),
-              nodeMap = state.nodeMap.updated(peer.id, Node(peer.id, address))
-            )
-          }
+          stateRef.update(_.withEnrAndAddress(peer, enr, address))
       }
 
     /** Forget everything about this peer. */
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
@@ -160,14 +160,23 @@ object DiscoveryService {
     def clearEnrFetch(peer: Peer[A]): State[A] =
       copy(fetchEnrMap = fetchEnrMap - peer)
 
-    def removePeer(peer: Peer[A]): State[A] =
-      copy(
-        nodeMap = nodeMap - peer.id,
-        enrMap = enrMap - peer.id,
+    def removePeer(peer: Peer[A], toAddress: Node.Address => A): State[A] = {
+      // We'll have ony one node/enr for this peer ID, but it may be with a different address.
+      // This can happen if we get a fake neighbor respose from a malicious peer, with the ID
+      // of an honest node and an ureachable address. We shouldn't remote the honest node.
+      (nodeMap.get(peer.id) match {
+        case Some(node) if toAddress(node.address) == peer.address =>
+          copy(
+            nodeMap = nodeMap - peer.id,
+            enrMap = enrMap - peer.id,
+            kBuckets = kBuckets.remove(peer.id)
+          )
+        case _ => this
+      }).copy(
         lastPongTimestampMap = lastPongTimestampMap - peer,
-        bondingResultsMap = bondingResultsMap - peer,
-        kBuckets = kBuckets.remove(peer.id)
+        bondingResultsMap = bondingResultsMap - peer
       )
+    }
 
     def removePeer(peerId: NodeId): State[A] =
       copy(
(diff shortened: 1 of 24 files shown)


scalanet/discovery/test/src/io/iohk/scalanet/discovery/ethereum/v4/mocks/MockSigAlg.scala
```diff
@@ -3,27 +3,44 @@ package io.iohk.scalanet.discovery.ethereum.v4.mocks
 import io.iohk.scalanet.discovery.crypto.{Signature, PublicKey, PrivateKey, SigAlg}
 import scodec.bits.BitVector
 import scodec.Attempt
+import scala.util.Random
 
 class MockSigAlg extends SigAlg {
   override val name = "MockSignature"
 
+  override val SignatureBytesSize = 65
+  // Normal Secp256k1 would be 33 bytes compressed or 64 bytes uncompressed.
```
KonradStaniec: this is comment is a littlee misleading. Compressed key is `32` bytes and uncompressed is `64` bytes. Both of this cases have always one byte pre-pended, and this byte indicate wheter key is compressed or uncompressed. Making final sizes `33` and `65` bytes.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -1,9 +1,798 @@
...
+              Task(logger.debug(s"Could not bond with $peer to fetch ENR")).as(None)
+            case true =>
+              Task(logger.debug(s"Fetching the ENR from $peer...")) >>
+                rpc
+                  .enrRequest(peer)(())
+                  .delayExecution(if (delay) config.requestTimeout else Duration.Zero)
+                  .flatMap {
+                    case None =>
+                      // At this point we are still bonded with the peer, so they think they can send us requests.
+                      // We just have to keep trying to get an ENR for them, until then we can't use them for routing.
```
KonradStaniec, line 575: just question, why succesfull enrRequest is required to use peer in routing ? Is it somewhere in spec or is based on other client ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -1,9 +1,798 @@
...
+            true
+          case _ =>
+            false
+        }
+        _ <- fetchEnr(peer, delay).whenA(needsFetching)
+      } yield ()
+
+    /** Fetch a fresh ENR from the peer and store it.
+      *
+      * Use delay=true if there's a high chance that the other side is still bonding with us.
```
KonradStaniec: when this condition can happen ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -1,9 +1,798 @@
...
+    * - maintain the state of K-buckets
+    * - return node candidates for the rest of the system
+    * - bond with the other nodes
+    * - respond to incoming requests
+    * - periodically try to discover new nodes
+    * - periodically ping nodes
+    */
+  def apply[A](
+      privateKey: PrivateKey,
+      node: Node,
```
KonradStaniec, line 68: Maybe we can validate that user provided info is really 65byte key ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -1,9 +1,798 @@
...
+    *
+    * - maintain the state of K-buckets
+    * - return node candidates for the rest of the system
+    * - bond with the other nodes
+    * - respond to incoming requests
+    * - periodically try to discover new nodes
+    * - periodically ping nodes
+    */
+  def apply[A](
+      privateKey: PrivateKey,
```
KonradStaniec, line 67: As i saw it is only used in `updateExternalAddress` maybe we can  pass it there as as function param i.e have 
` override def updateExternalAddress(ip: InetAddress, privateKey: PrivateKey)`. That way user of library would not be forced to keep private key in memory during whole time.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -1,9 +1,798 @@
...
+    /** Runs the bonding process with the peer, unless already bonded.
+      *
+      * If the process is already running it waits for the result of that,
+      * it doesn't send another ping.
+      *
+      * If the peer responds it waits for a potential ping to arrive from them,
+      * so we can have some reassurance that the peer is also bonded with us
+      * and will not ignore our messages.
+      */
+    protected[v4] def bond(
```
KonradStaniec, line 394: from what i understand bonding is `ping - pong - fetch_enr` sequence ?

KonradStaniec (commented):
rather minor comments and questions. In my opinion it is  litte hard to follow what is happening and when, but this is probably non-avoidable complexity of this whole protocol. 
Imo maybe it would be probably the best to integrate it as soon as possible to mantis or our other project, and catch most of the problems on running cluster

scalanet/discovery/test/src/io/iohk/scalanet/discovery/ethereum/v4/mocks/MockSigAlg.scala
aakoshh: Hm, you're right, I knew I remembered it being 65 bytes long. I got confused by [this](https://ethereum.stackexchange.com/questions/1177/what-is-an-enode-id-in-ethereum) answer saying it's 512 bits long.

Just seen now that the ENR signature wants to omit the recovery byte, making it just 64 bytes long :(

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: The spec says this:

> To resolve the current record of any node public key, perform a Kademlia lookup using FindNode packets. When the node is found, send ENRRequest to it and return the record from the response.

It was added in [EIP-8](https://eips.ethereum.org/EIPS/eip-868) but I'm surprised it's not older.

That sounds like something you'd use in routing, when you have the identity of a node but not its address yet. We don't use it like that though. 

My reasoning was that a malicious peer could return `Node` records with altered TCP ports for example. We have the requirement that we have to bond with peers before we send them messages, so during recursive `lookup` we have to Ping them, which gives us some proof that they a) respond to the UDP port and b) the public key they used to sign their response matched the node ID we expected them to have; however we still cannot be 100% sure what their TCP port is, unless we fetch their ENR record (which they also signed, although they sign the packet that transmits it as well).

Trinity has them but it's a bit confusing as they insert a "stub" version based on the packet and then later try and resolve it fully. It's also persistent, survives restarts.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: It happened in your comprehensive integration tests. 
When Alice wants to talk to Bob, she has to bond with him first, and then Bob has to bond with Alice as well.
There's a sequence of events that happened all the time:

1. Alice wants to bond with Bob: Alice sends Ping to Bob
2. Bob receives the Alice's Ping and initiates an asynchronous bond as well: Bob sends a Ping to Alice.
3. Bob responds to Alice's Ping: Bob sends a Pong to Alice (may go before the Ping).
4. Alice receives Bob's Pong. She considers them bonded now and initiates an async ENR fetch: Alice sends ENRRequest to Bob.
5. Bob receives Alice's ENRRequest but he hasn't got a Pong yet so ignores it, they aren't yet bonded as far as he's concerned.
6. Alice receives Bob's Ping: Alice sends Pong to Bob. Maybe she also initiates an async ENR fetch.
7. Bob receives Alice's Pong. Now he considers her bonded. This may happen after the ENRRequest arrives.

Due to the way the `DiscoveryRPC` works we should be careful of sending messages from the handlers themselves: they may arrive earlier than the response given by the handler. For example in the `ping` handler we can start a `fetchEnr` but that may result in a message before the result of the `ping` invocation, the `Pong`, reaches the peer. While this isn't ideal, since this is UDP, we wouldn't get any ordering guarantee anyway, so if we quickly do a `send(Pong); send(ENRRequest)` they still may end up arriving in a different order. Therefore I figured the simplest way to avoid the problem is to introduce a little delay in situations when we expect that the other party may be pinging us, waiting for our pong.

Another option would be to do retries. I just did this to make the integration tests pass.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: I'd add validation to data we ingest from other nodes, but to validate the input parameters here would beg the question: why here, why not somewhere else, like when the key was deserialized from disk? We could end up with such validations littered absolutely everywhere and then it becomes a bit paranoid, don't you think?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Well the `DiscoveryNetwork` also needs a copy of the private key since it has to sign every outgoing packet. Maybe you'd be happier with a `privateKeyReader: Task[PrivateKey]` that may or may not load the key, on demand? 

The reason I hesitate to add it to the `updateExternalAddress` is that it's a method that looks like would be called by a pure networking component that somehow learns about our external IP: why should it know the key, or the fact that this component needs to sign an ENR?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Strictly speaking it just tries to achieve a mutual ping, so want to have reasonable confidence that this happened, if it had to:
```
Ping -> 
        <- Pong
        <- Ping
Pong ->
```
It's possible that the other side didn't need to Ping us, because we have responded earlier to a Pong, in the last 12 hours.

scalanet/discovery/test/src/io/iohk/scalanet/discovery/ethereum/v4/mocks/MockSigAlg.scala
aakoshh: Fixed it.

aakoshh:
Thanks for the review, great points!

> In my opinion it is litte hard to follow what is happening and when, but this is probably non-avoidable complexity of this whole protocol.

I agree, it's difficult to reason about the state because both sides can forget any time what happened earlier, so there's a lot of "just in case" in there. The [v5 rationale](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-rationale.md) enumerates many shortcomings with the v4 protocol. Node records will be gone in favor of ENRs, endproof verification replaced with handshake, etc.

And yeah, it would be difficult to work on this any longer without implementing the missing RLP and crypto bits, so it needs to connect to Mantis very soon.

aakoshh:
The summary is that before sending any request to a peer, we have to call `bond`, which checks if 1) the last ping has been too long ago, in which case it pings now and 2) if we have an ENR from the peer, initiating an async fetch if we don't. `lookup` is only ever used to collect peers, and as a side effect kick off a bunch of background ENR fetches. Eventually some of these will end up populating the node map which is where peer candidates are returned when the host application asks.

scalanet/discovery/test/src/io/iohk/scalanet/discovery/ethereum/v4/mocks/MockSigAlg.scala
aakoshh: Ah wait, all node IDs are expected to be 64 bytes long according to the [lookup in trinity](https://github.com/ethereum/trinity/blob/dc0ddbb5ee5129b29ea2d4a9444f6a1b60cf09f4/p2p/discovery.py#L580), the prefix is omitted.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Added the checks here and in `lookup` as well.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
KonradStaniec: but even the docs of this method says `incrementing the local ENR sequence.` which kinda implies sigining enr sequence with new number. 

To be honest from mantis perspective it is not that important for now, as we are keeping private key in memory there eitherway.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
KonradStaniec: ok, so lets leave it as it is and we will check how it works in practice

KonradStaniec (approved):

PR #97, opened 2020-10-20
ETCM-254: Allow packet size violation

aakoshh:
Noticed that [besu](https://github.com/hyperledger/besu/blob/master/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java#L555-L556) sends all nodes in `Neighbors` that it finds, rather than just what fits into 1280 bytes as per the v4 protocol specification. To be compatible, the PR adds an optional setting to allow incoming packets to violate the size limit. `Packet.packetCodec` is no longer implicit, we have to tell it whether violations are allowed. 

The PR also adds `StaticUDPPeerGroup.Config.receiveBufferSizeBytes` which can be used to set the incoming buffer size to a lower value than the default 64KiB used by `DefaultMaxBytesRecvByteBufAllocator`. It looks like [NioDatagramChannel.doReadMessages](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/socket/nio/NioDatagramChannel.java#L241-L270) only ever reads one message at a time, so by allocating a 64KiB buffer for each message we were wasting a bit of resources. Limiting it to, say, 2*1280 like Trinity would mean that small packets like `Ping` wouldn't end up in a large buffer they don't need, and if some client tries to send much more than 1280 bytes they stop at Netty.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
@@ -225,14 +225,26 @@ class StaticUDPPeerGroup[M] private (
         success
     }
 
+  private def bufferAllocator: RecvByteBufAllocator = {
+    // `NioDatagramChannel.doReadMessages` allocates a new buffer for each read and
+    // only reads one message at a time. UDP messages are independent, so if we know
+    // our packages have a limited size (lower than the maximum 64KiB supported by UDP)
+    // then we can save some resources by not over-allocating and also protecting
+    // ourselves from malicious clients sending more than we'd accept.
+    val maxBufferSize = 64 * 1024
+
+    val bufferSize =
+      if (config.receiveBufferSizeBytes <= 0) maxBufferSize
+      else math.min(config.receiveBufferSizeBytes, maxBufferSize)
+
+    new io.netty.channel.FixedRecvByteBufAllocator(bufferSize)
+  }
+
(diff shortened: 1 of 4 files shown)


scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
```diff
@@ -225,14 +225,26 @@ class StaticUDPPeerGroup[M] private (
         success
     }
 
+  private def bufferAllocator: RecvByteBufAllocator = {
```
KonradStaniec, line 228: couldn't it be `val` or `lazy val` to avoid constant creating of allocator object ?

KonradStaniec (approved):
just a minor comment

scalanet/src/io/iohk/scalanet/peergroup/udp/StaticUDPPeerGroup.scala
aakoshh: Could be, but it's only ever created once as part of the `lazy val serverBinding`, I just wanted to move out out of there so I can add some comments.

aakoshh: Internally it will also just use a single handle instance, unlike `DefaultMaxBytesRecvByteBufAllocator` which does allocate a new one for each message.

PR #102, opened 2020-11-06
ETCM-168: Fix stalled enrollments

aakoshh:
Sometimes `enroll` seems to never finish. After debugging I think it's the `bondNeighbors`.

I couldn't see anything obvious in the code, but it's conceivable that if there are multiple bonding attempts then a deferred somehow got stuck in the map. Instead of waiting on it forever it now returns a result after a timeout.

Also limited the number of parallel bonding attempts to `alpha`.

### DIFF
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
@@ -797,7 +797,7 @@ object DiscoveryService {
           ) >>
             Task
               .parTraverseUnordered(contacts)(fetchNeighbors)
-              .map(_.flatten.distinct)
+              .map(_.flatten.distinct.filterNot(neighbors))
               .flatMap(bondNeighbors)
               .flatMap { newNeighbors =>
                 val nextClosest = (closest ++ newNeighbors).take(config.kademliaBucketSize)
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
@@ -407,7 +407,7 @@ object DiscoveryService {
         case false =>
           initBond(peer).flatMap {
             case Some(result) =>
-              result.pongReceived.get
+              result.pongReceived.get.timeoutTo(config.requestTimeout, Task.pure(false))
 
             case None =>
               Task(logger.debug(s"Trying to bond with $peer...")) >>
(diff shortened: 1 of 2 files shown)


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -775,18 +786,19 @@ object DiscoveryService {
...
-            .flatMap(bondNeighbors)
-            .flatMap { newNeighbors =>
-              val newClosest = (closest ++ newNeighbors).take(config.kademliaBucketSize)
-              val newAsked = asked ++ contacts
-              loop(local, newClosest, newAsked)
-            }
+        if (contacts.isEmpty) {
+          Task(logger.debug(s"Lookup finished for $target after asking ${asked.size} nodes.")).as(closest)
+        } else {
+          Task(logger.debug(s"Lookup for $target contacting ${contacts.size} nodes.")) >>
```
KonradStaniec: Maybe we could also log how many nodes we already asked ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -775,18 +786,19 @@ object DiscoveryService {
...
+            Task
+              .parTraverseUnordered(contacts)(fetchNeighbors)
+              .map(_.flatten.distinct)
+              .flatMap(bondNeighbors)
+              .flatMap { newNeighbors =>
+                val newClosest = (closest ++ newNeighbors).take(config.kademliaBucketSize)
```
KonradStaniec: Maybe we could log how many new neighbours we had found ?

KonradStaniec (approved):

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Sure thing.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: :+1:

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: I thought we could log the addresses too but it's a bit verbose.

PR #103, opened 2020-11-06
ETCM-168: De-restrict parallelism to speed up enrollment.

aakoshh:
Enable unrestricted parallelism when enrolling with bootstraps and bonding with neighbors to speed up the initialisation.

Also fixes the CI build config to run unit and integration tests separately. `mill __.test` was actually matching `mill scalanet.test.test` as well as `mill scalanet.discovery.it.test`, so it ran both.


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -762,7 +762,7 @@ object DiscoveryService {
         for {
           _ <- Task(logger.debug(s"Bonding with ${neighbors.size} neighbors..."))
           bonded <- Task
-            .parTraverseN(config.kademliaAlpha)(neighbors) { neighbor =>
+            .parTraverseUnordered(neighbors) { neighbor =>
```
KonradStaniec, line 772: doesn't that mean that if malicious node send us lets say 1000 neighbours, we will start 1000 parallel bonding tries and it could potentially exhaust our resources ?

aakoshh: They can try to send us 1000 nodes, but a single `Neighbors` message can only contain ~10 nodes, and we'll anyway [discard](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala#L274) anything beyond the first 16, so these will just be dropped. 

What I thought is that we don't have that many resources tried up in bonding after all: we use the same single UDP socket to send and receive, so it's just how rapidly we have to react to incoming messages. But the impact of non-responding nodes with their 3 second timeout is really visible in mordor, takes minutes to get past the initialisation.

KonradStaniec: hmm fair enough. Maybe one more question, if the enrollement really takes such long time on mordor (and on mainner it will probably take much longer), maybe we should start enrollment in background and return from `DiscoveryService.apply` immediately ? Then caller will be able to get already found peers, event hough enrollment did not finish yet. ( of course then we would need to start random lookups only when enrollment finishes, to avoid several concurrent lookups in progress)

aakoshh: We can do that yeah, I tried to follow the `KRouter` which waits until enrollment is done, however the semantics of enrollment are not black and white: even if it fails during initialisation, it will be re-attempted during lookups unless we have at least 16 closest nodes already to the target, so in case we have just 1 bootstrap node and it happens to be offline, we can still connect to it later. And yes, as you say the caller would be able to read discovered nodes earlier, before the whole enrollment is finished.

In Mantis this wouldn't be a major issue though because the default setting is to consider bootstrap nodes and previously persisted nodes as "already discovered" and serve them, while the enrollment runs in the background. 

At the very least the initial self-lookup could be done in the background since subsequent lookups are also happening outside the control of the host application.

aakoshh: Added option to run in background in https://github.com/input-output-hk/scalanet/pull/103/commits/cf7358cd011fdcc3a5873910076731c63fc8eb41 with the default left as foreground because that's what integration tests assume.

KonradStaniec (approved):

PR #106, opened 2020-11-10
ETCM-199: Remove ConnectableSubject

aakoshh:
The PR removes the `ConnectableSubject`, which means `PeerGroup.server` and `Channel.in` are gone and it's no longer possible to set up subscribers then call `.connect()`. This should simplify the semantics we have to keep in our head and things like the test I had to disable for ETCM-345, now re-enabled, are easier to reason about.

To help with the migration, `import io.iohk.scalanet.peergroup.implicits._` brings into scope some extension methods that can help:
* `peerGroup.server.refCount` => `peerGroup.serverEventObservable`
* `channel.in.refCount` => `channel.channelEventObservable`

The `channel.nextMessage` method got renamed to `channel.nextChannelEvent` with the corresponding `channel.channelEventObservable`. It was named "nextMessage" because of earlier discussions where we thought about names like "readMessage" due to the proximity of `channel.sendMessage`, but I thought since there are other non-message members of the ADT it's more consistent to match the `PeerGroup` in this case.

The `ReqResponseProtocol` has an example of dealing the situation where multiple consumers need to be attached to the same channel, which the `.in.refCount` supported, with the caveat that if the subscriber count went down to 0 it would unsubscribe from the source (I thought permanently, but I'm not sure any more). In our case subscriptions come and go with each request, while the channel stays open, so the options were to either use `val obs = channel.channelEventObservable.share` and start a `obs.foreach(_ => ())` in the background to keep it alive and to discard any unwanted incoming message, or as the code is now use a `monix.catnap.ConcurrentChannel` which supports broadcasting multiple consumers.

The version is increased to `0.5.0` because it's a breaking change.


aakoshh:
@KonradStaniec I have a stylistic question: 

I used parentheses in the name of methods such as `PeerGroup.nextServerEvent()` or `CloseableQueue.closeAndDiscard()` because they have an effect on the object they are called on, unlike for example `PeerGroup.processAddress` which just reads, following the basic Scala convention of marking methods without parameters but with side effects with `()`. I thought these draw our attention to the fact that calling `next()` on the queue consumes an item from it, which won't be available for other consumers. However these methods actually don't have side effects when they are called, since they return a `Task`, only when the task is executed does the effect take place. I thought this was an implementation detail, if it returned a `Future` or if it was synchronous then the `()` would be justified. On the other hand for example `monix.catnap.ConcurrentQueue` doesn't use them for stuff like `.poll`. 

What do you think, should we drop them?

EDIT: I also looked at Doobie, methods like `.close` don't have parentheses, so I ended up removing them.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -48,8 +48,6 @@ class DynamicTLSPeerGroup[M] private (val config: Config)(
 ) extends TerminalPeerGroup[PeerInfo, M]
     with StrictLogging {
 
-  protected override val s = scheduler
-
   private val sslServerCtx: SslContext = DynamicTLSPeerGroupUtils.buildCustomSSlContext(SSLContextForServer, config)
 
   private val serverQueue = CloseableQueue.unbounded[ServerEvent[PeerInfo, M]](ChannelType.SPMC).runSyncUnsafe()
```
KonradStaniec: I am wondering, that if maybe with currrent setup we should have this queue as `ChannelType.MPMC` ? From what i can see all events to it are pushed from `implicit scheduler: Scheduler` provided by user to class, and there is no guarannte that user of the class will provide one threaded scheduler.

Other solution (more neat in my opinion) would be to create one threaded `NioEventLoopGroup` for accepting connections and pushing sever events on `serverQueue`, then this external scheduler could be deleted completely and all producing would happen on netty threads. Then you would have :
```
  private val workerGroup = new NioEventLoopGroup()
  private val acceptinGroup = new NioEventLoopGroup(1)
  
  private val serverBootstrap = new ServerBootstrap()
    .group(acceptinGroup, workerGroup)
```

wdyt ?

KonradStaniec:
and about stylistic question, I do not have strong opinion on that to be hones. If doobie goes without parentheses, then we can also go that way.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: My current understanding is that `SPMC` is fine because of what I pointed to in this comment: https://github.com/input-output-hk/scalanet/pull/92#discussion_r501611101

Basically I think that unless you actually publish from multiple threads at the same time, you are fine with the "single producer" scenario, which in our case is true because:
* we [push server events](https://github.com/input-output-hk/scalanet/blob/717cc24572a58f7548a7af0a4745791ac2373100/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala#L267) with `.runSyncUnsafe()` which is running on the external scheduler but the thread accepting netty connections one will wait for it, so it won't try to offer anything else at the same time
* we [push messages](https://github.com/input-output-hk/scalanet/blob/717cc24572a58f7548a7af0a4745791ac2373100/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala#L45-L46) on a single threaded scheduler on top of the netty one as you suggested, so again no chance to do two at the same time

I actually started to remove these schedulers, but then gave up when I realised they are present mainly so that we can block creating the queue instance. Ideally IMO we could refactor the TLS group to look more like the `StaticUDPPeerGroup` and do almost everything in tasks, async, on top of the netty channel.

KonradStaniec: this the case i am worried about:

> we push server events with .runSyncUnsafe() which is running on the external scheduler but the thread accepting netty connections one will wait for it, so it won't try to offer anything else at the same time

as each connection currently is accepted on possibly different thread. So there can be potentially mulitple handshakes in progress, an if lets say two of them end up in the same time then both can run `.runSyncUnsafe()` and if external scheduler is muliple threaded then we will have multiple producers.

aakoshh: Hm, I thought Netty only has 1 boss thread where it accepts connections. I see that we use the same work group for accepting connections and handling messages, but but still I thought it will always the be same thread doing a particular task.

aakoshh: https://stackoverflow.com/questions/5474372/how-netty-uses-thread-pools says there's only one boss thread per socket channel.

KonradStaniec: Hm yep this thread may be right, but when we talk about accepting connection we are talking about tcp handshake. When we are push our success event:
https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala#L249

we are way past that point. This is happening on each channel thread, and there can be many muliple tls handshakes (incoming channels) at any time .
At  this point I am pretty convinced that split ` .group(acceptinGroup, workerGroup)` would not help, as it is still at accepting tcp connection level, and only solution is to either to guarantee that scheduler used to push events on `serverQueue` is single threaded, or make this queue MPMC.

aakoshh: That's a good point, I didn't realise that, thanks! 

Changed it to MPMC.

KonradStaniec (approved):

PR #107, opened 2020-11-19
ETCM-393: Expose the lookup methods on DiscoveryService

aakoshh:
Allow `lookup` and `lookupRandom` to be called on the `DiscoveryService` so Mantis can trigger searches on demand, rather than wait on the fixed schedule to yield more peers to connect to.

### DIFF
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
@@ -42,6 +42,12 @@ trait DiscoveryService {
 
   /** The local node representation. */
   def getLocalNode: Task[Node]
+
+  /** Lookup the nodes closest to a given target. */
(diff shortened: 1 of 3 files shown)


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -42,6 +42,12 @@ trait DiscoveryService {
 
   /** The local node representation. */
   def getLocalNode: Task[Node]
+
+  /** Lookup the nodes closest to a given target. */
+  def lookup(target: Node.Id): Task[SortedSet[Node]]
+
+  /** Lookup a random target, to discovery new nodes along the way. */
+  def lookupRandom: Task[Set[Node]]
```
aakoshh, line 50: Removed `()` like in https://github.com/input-output-hk/scalanet/pull/106

KonradStaniec (approved):

PR #109, opened 2020-11-26
ETCM-416: Limit entries from the same subnet in the K-table

aakoshh:
The PR implements what `geth` does to limit the entries in the K-table from the same subnet, which is a variation of _Countermeasure 2_ from the [eclipse attack paper](https://eprint.iacr.org/2018/236.pdf). In particular, by default it will not add a peer to the k-table if doing so would violate the limits. The default limits are that there cannot be more than 2 peers from the same subnet in any k-bucket, or more than 10 across all buckets, where the subnet is defined as the fist 24 bits of the IP address. All these values are configurable.

Note that the nodes and ENR records are still saved by the service, the only difference is that these values will not be inserted into the k-table, which should prevent an attacker from filling up the k-table with nodes running on the same machine, or the same subnet, thus hijacking all `FindNode` requests, making our node unable to discover honest peers. Because of this, Mantis should also have some measures to limit the number of TCP connections going to the same IP or subnet, because if it uses `getNodes` it will still get every discovered peer.


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -228,7 +229,13 @@ object DiscoveryService {
       )
     }
 
-    def removePeer(peerId: Node.Id): State[A] =
+    def removePeer(peerId: Node.Id, toAddress: Node.Address => A): State[A] = {
+      // Find any Peer records that correspond to this ID.
+      val peers =
+        nodeMap.get(peerId).map(node => Peer(node.id, toAddress(node.address))).toSeq ++
+          lastPongTimestampMap.keys.filter(_.id == peerId).toSeq ++
```
KonradStaniec, line 236: seems wasteful that we are itereating `lastPongTimestampMap` and `bondingResultsMap` maps two times here when finding all records of peer, and later when fitering this peer out. From what i known this collections can be fairly  large, maybe we could do it in one pass ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/KBucketsWithSubnetLimits.scala
```diff
@@ -0,0 +1,106 @@
...
+    if (contains(peer)) this
+    else {
+      val ip = subnet(peer)
+      val idx = getBucket(peer)._1
+
+      val tlc = tableLevelCounts |+| Map(ip -> 1)
+      val blc = bucketLevelCounts |+| Map(idx -> Map(ip -> 1))
+
+      val overTheLimit =
+        limits.prefixLength > 0 && (
```
KonradStaniec: maybe we could add  some comments about each case or export this to separate funcion with some docs as this logic is little dense ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/KBucketsWithSubnetLimits.scala
```diff
@@ -0,0 +1,106 @@
...
+    if (!contains(peer)) this
+    else {
+      val ip = subnet(peer)
+      val idx = getBucket(peer)._1
+
+      val tlc = tableLevelCounts |+| Map(ip -> -1) match {
+        case counts if counts(ip) <= 0 => counts - ip
+        case counts => counts
+      }
+      val blc = bucketLevelCounts |+| Map(idx -> Map(ip -> -1)) match {
```
KonradStaniec: maybe we could add some comments about each case or export this to separate funcion with some docs as this logic is little dense ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Sure, I can change that. But this method won't be actually called, I just kept it for backwards compatibility with KRouter.

aakoshh: Done.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/KBucketsWithSubnetLimits.scala
aakoshh: Done, split it out into several methods.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/KBucketsWithSubnetLimits.scala
aakoshh: Done, split it out into several methods.

KonradStaniec (approved):

PR #111, opened 2020-12-15
[FIX] Fix missing arg

KonradStaniec:
Use proper arg when using `ClosableQueue.unbounded` constructor


aakoshh (approved):
:eagle: :eyes: !

PR #112, opened 2020-12-20
Scala 2.13 cross-build upgrade

dmitry-worker:
Scala 2.13 port of `scalanet` is a blocker to `Mantis` backlog task.

A library can be built now against both - `2.12` and `2.13` targets.
Further advice is needed regarding documentation and publishment.


build.sc
```diff
@@ -6,147 +6,164 @@ import coursier.maven.MavenRepository
...
-  override def scalaVersion = "2.12.10"
+object csm extends Cross[ScalanetModule]("2.12.10", "2.13.4")
 
-  override def repositories =
-    super.repositories ++ Seq(
-      MavenRepository("https://oss.sonatype.org/content/repositories/releases"),
-      MavenRepository("https://oss.sonatype.org/content/repositories/snapshots")
-    )
+class ScalanetModule(crossVersion: String) extends Module {
+  override def millSourcePath = super.millSourcePath / os.up / os.up
```
aakoshh, line 26: Could you add comments to why this is needed now?

build.sc
```diff
@@ -6,147 +6,164 @@ import coursier.maven.MavenRepository
...
 
-  private val commonScalacOptions = Seq(
-    "-unchecked",
-    "-language:postfixOps",
-    "-deprecation",
-    "-feature",
-    "-Xfatal-warnings",
+  private val pointTwelveOptions: Seq[String] = Seq(
```
aakoshh: Until I saw `pointThirteenOptions` I thought you were talking about [the 12 factor-app](https://12factor.net/). Maybe 
```
private val `scala 2.12 options`: Seq[String] = ???
``` 

Can you add some comment on why these options are needed in one version but not the other?

build.sc
```diff
@@ -6,147 +6,164 @@ import coursier.maven.MavenRepository
...
-      ivy"ch.qos.logback:logback-core:1.2.3",
-      ivy"ch.qos.logback:logback-classic:1.2.3",
-      ivy"org.mockito:mockito-core:2.21.0"
-    )
+    // `extends Tests` uses the context of the module in which it's defined,
+    // which is why the trait is defined here not within `scalanet`, otherwise
+    // it wouldn't work for `kademlia` for example.
+    trait TestModule extends Tests {
+      override def scalacOptions = Nil
+        //commonScalacOptions
```
aakoshh: Can it be `Nil`? Does it inherit something? What about the commented out `commonScalacOptions`?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/codecs/DefaultCodecs.scala
```diff
@@ -38,6 +39,21 @@ object DefaultCodecs {
...
+      while (ai.hasNext && bi.hasNext && result.isEmpty) { 
+        val res = Byte.compare(ai.next(), bi.next())
+        if (res != 0) result = Some(res)
+      }
+      result.getOrElse( Boolean.compare(ai.hasNext, bi.hasNext) )
+    }
+  }
+
+  implicit val byteIterableOrdering = new ByteIterableOrdering
   implicit val byteVectorOrdering: Ordering[ByteVector] =
```
aakoshh: Did you need to add it because in 2.13 it's [deprecated](https://github.com/scala/scala/blame/v2.13.0/src/library/scala/math/Ordering.scala#L530-L536)? If so, can you add some comments? 

It looks though that the [Ordering.Implicits](https://github.com/scala/scala/blob/2.12.x/src/library/scala/math/Ordering.scala#L229-L230) they point us to also exists in 2.12 and should work for `Seq`, so probably we should import [seqDerivedOrdering](https://www.scala-lang.org/api/2.12.1/scala/math/Ordering$$Implicits$.html#seqDerivedOrdering[CC[X]%3C:Seq[X],T](implicitord:scala.math.Ordering[T]):scala.math.Ordering[CC[T]]) instead of copying this code.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -193,11 +193,11 @@ object DiscoveryNetwork {
         maybeResponse.attempt.flatMap {
           case Right(Some(response)) =>
             f(response)
-          case Right(None) =>
-            Task.unit
           case Left(NonFatal(ex)) =>
             // Not responding to this one, but it shouldn't stop handling further requests.
             Task(logger.error(s"Error handling incoming request: $ex"))
+          case _ => // Right(None) or fatal? - exhaustive match
+            Task.unit
```
aakoshh: Hm, this ain't right. Think about it: a `Right(None)` means no response which is normal, and I wanted to log unexpected errors, but not catch fatal ones. This change would silently swallow fatal errors like stack overflow or OOM. 

I suggest writing it like this instead:

```scala
maybeResponse.flatMap(_.fold(Task.unit)(f)).onErrorRecoverWith {
  case NonFatal(ex) =>
    // Not responding to this one, but it shouldn't stop handling further requests.
    Task(logger.error(s"Error handling incoming request: $ex"))
}
```

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -385,6 +385,10 @@ object DiscoveryNetwork {
                 // New response, fold it with the existing to decide if we need more.
                 val next = (acc: Z) => Some(acc -> (count + 1))
                 Task.pure(f(acc, response).bimap(next, next))
+
+              case _ => 
+                // Invalid state - this cannot happen
+                Task(logger.debug(s"Unexpected state while collecting responses from ${channel.to}")).as(Right(None))
```
aakoshh:
```suggestion
              case (None, _) => 
                Task.raiseError(new IllegalStateException(s"Unexpected state while collecting responses from ${channel.to}"))
```
So it doesn't go unnoticed.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
```diff
@@ -10,11 +10,11 @@ import io.netty.handler.ssl.{ClientAuth, SslContext, SslContextBuilder}
 import javax.net.ssl._
 import scodec.bits.BitVector
 
-import scala.collection.JavaConverters._
+import java.util.Arrays
 
 private[scalanet] object DynamicTLSPeerGroupUtils {
   // supporting only ciphers which are used in TLS 1.3
-  val supportedCipherSuites: Seq[String] = Seq(
+  val supportedCipherSuites: java.lang.Iterable[String] = Arrays.asList(
```
aakoshh:
```suggestion
  private val supportedCipherSuites: java.lang.Iterable[String] = Arrays.asList(
```
If it's optimised for Java interop it's better to limit its visibility.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/EthereumNodeRecord.scala
```diff
@@ -16,8 +17,8 @@ case class EthereumNodeRecord(
 )
 
 object EthereumNodeRecord {
-  implicit val byteOrdering: Ordering[ByteVector] =
-    Ordering.by[ByteVector, Iterable[Byte]](_.toIterable)
+
+  implicit val byteOrdering: Ordering[ByteVector] = DefaultCodecs.byteVectorOrdering
```
aakoshh: I was really trying to keep `DefaultCodecs` out of the code so that it's imported _only_ if someone wants to use scodec, and only at the point where the object is instantiated, so for example Mantis where we use RLP doesn't need to touch these at all. I suggest, as per my other comment, that we import `Ordering.Implicits.seqDerivedOrdering` here, or if that doesn't work than separate out ordering related utils from codecs.

build.sc
```diff
@@ -6,147 +6,164 @@ import coursier.maven.MavenRepository
...
-    ivy"org.bouncycastle:bctls-jdk15on:1.64"
-  )
+  // In objects inheriting this trait, use `override def moduleDeps: Seq[PublishModule]`
+  // to point at other modules that also get published. In other cases such as tests
+  // it can be `override def moduleDeps: Seq[JavaModule]`.
+  trait ScalanetPublishModule extends PublishModule {
+    def description: String
 
-  def scoverageVersion = "1.3.1"
+    override def publishVersion = "0.5.0-SNAPSHOT"
```
aakoshh:
```suggestion
    override def publishVersion = "0.5.1-SNAPSHOT"
```
It's a bit of a pain but once we start using a version in a project which uses Nix expressions we have to increment the version here otherwise all builds in those projects will constantly break down because of hash changes. Konrad is already migrating to 0.5.0 so we should bump it. 

I guess this will be difficult to coordinate, so we should have explicit release process which removes the `-SNAPSHOT`, and projects that _do_ want to compile against SNAPSHOT should deal with the Nix expressions. @KonradStaniec how do you do releases?

aakoshh (commented):
Thank you for figuring out the cross build, good job!

Are all commands the same, does it automatically publish both versions?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
aakoshh: Although that's not the same either since it would also handle any errors raised by `f`, which would practically mean that the channel could not send and should therefore let the error bubble up and cause it to be released. 

Another option would be to add 
```scala
  case ex => 
     Task.raiseError(ex)
```
and call it a day.

aakoshh: Last variant :) 
```scala
maybeResponse.onErrorRecoverWith {
  case NonFatal(ex) =>
    // Not responding to this one, but it shouldn't stop handling further requests.
    Task(logger.error(s"Error handling incoming request: $ex")).as(None)
}.flatMap(_.fold(Task.unit)(f))
```

dmitry-worker:
> Thank you for figuring out the cross build, good job!

Thanks :) I hope my effort will make this project even better.

> Are all commands the same, does it automatically publish both versions?

No, the pull request is still draft, hope I can finish the work in 2-3 days.
Documentation update is not ready as well

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/EthereumNodeRecord.scala
dmitry-worker: It's done!

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
dmitry-worker: Oops, I have missed that temporary `let-it-compile` fix.
Will fix that like you suggested last.

build.sc
dmitry-worker: My bad, fixed that.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/codecs/DefaultCodecs.scala
dmitry-worker: Unfortunately I had to add the implementation because 2.12 and 2.13 orderings came from different packages. 
I will continue researching the possibility of ordering without triggering deprecation warnings.
I have also moved this monster to `util`ity package.

build.sc
dmitry-worker: 0.5.1 is now officially set :1st_place_medal:

build.sc
dmitry-worker: Done, commented in build.sc

build.sc
dmitry-worker: Done, commented in build.sc

scalanet/ut/src/io/iohk/scalanet/codec/FramingCodecSpec.scala
```diff
@@ -23,7 +23,7 @@ class FramingCodecSpec extends FlatSpec with EitherValues {
 
     forAll(genString()) { listOfString =>
       val enc = listOfString.map(encoder.encode(_).require).foldLeft(BitVector.empty)((acc, bit) => acc ++ bit)
-      val decoded = encoder.streamDecode(enc).right.value
+      val decoded = encoder.streamDecode(enc).toOption.get
```
dmitry-worker: Sorry for that (and for `.swap.toOption.get` as well) :(
I found no alternatives to fit both 2.12 and 2.13 `Either` methods without triggering any compiler warning.

scalanet/discovery/src/io/iohk/scalanet/kademlia/TimeSet.scala
```diff
@@ -28,6 +28,9 @@ class TimeSet[T] private (val clock: Clock, val timestamps: HashMap[T, Long], va
   override def iterator: Iterator[T] =
     underlying.iterator
 
+  def diff(that: Set[T]): Set[T] = 
+    underlying &~ that
```
dmitry-worker, line 32: A `diff` method arrived only in 2.13

build.sc
KonradStaniec: @aakoshh to this time i have only done release one time so the process is a bit ad hoc and manual:
1.  create release branch
2. create commit which removes `-SNAPHOST` version, and add necessary modification to build.sc (last time I needed to modify it add myslef do `developers` field to by able to do realease). In general we do not have mill version, and thigs change based on mill version, which breaks things pretty often. (maybe we should fix mill version to make it easier, wdyt ?)
3. tag this commit
4. run mill incantation i.e the same thing which is running when merging to develop, but with `--relase` flag

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -385,6 +381,10 @@ object DiscoveryNetwork {
                 // New response, fold it with the existing to decide if we need more.
                 val next = (acc: Z) => Some(acc -> (count + 1))
                 Task.pure(f(acc, response).bimap(next, next))
+
+              case _ =>
```
aakoshh:
```suggestion
              case (None, _) => 
```
Makes it easier to understand why.

scalanet/discovery/src/io/iohk/scalanet/discovery/util/package.scala
```diff
@@ -0,0 +1,27 @@
+package io.iohk.scalanet.discovery
+
+import scodec.bits.ByteVector
+import scala.math.Ordering._
+
+
+package object util { 
+
+  private class ByteIterableOrdering extends Ordering[Iterable[Byte]] {
```
aakoshh: Can this not be used? 

https://www.scala-lang.org/api/2.13.0/scala/math/Ordering$$Implicits$.html#seqOrdering[CC[X]%3C:scala.collection.Seq[X],T](implicitord:scala.math.Ordering[T]):scala.math.Ordering[CC[T]]

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
```diff
@@ -6,15 +6,22 @@ import java.security.KeyStore
...
-    trustStore.aliases().asScala.toArray.map(alias => trustStore.getCertificate(alias))
+  val trustedCerts: Array[Certificate] = { 
+    val aliases = trustStore.aliases()
+    val buffer = ArrayBuffer[Certificate]()
+    while(aliases.hasMoreElements()) {
+      val alias = aliases.nextElement()
+      buffer.append(trustStore.getCertificate(alias))
+    }
+    buffer.toArray
+  }
```
aakoshh: I think this is not a good tradeoff for something that runs once.

scalanet/ut/src/io/iohk/scalanet/codec/FramingCodecSpec.scala
aakoshh: What was the problem here?

dmitry-worker: Calling `.right` on `Either` produces a deprecation warning: 
In scala 2.13 `Either` has become right-based.

We could probably disable fatal-warnings directive but it will make the codebase even worse.
I wouldn't try doing some ad-hoc methods for `Either`
Or even separate sources for 2.12 and 2.13.

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
aakoshh: I mean in general it's harder to understand this kind of Javaesque looping. It might be worth doing if the code turns out to be a hotspot, or maybe if it's hidden away to a library method, but here inline for some field that gets initialized once, I think the `.asScala` is a pretty idiomatic way to handle the conversion and the 1 line we had was easy to understand. It actually seems to return an `Iterator[String]` in this case, so perhaps moving `.toArray` to the end would avoid the interim allocation that I assume you're worried about. 

However this change has nothing to do with the PR, does it? Or has something changed in 2.13? If not, I think it's best to concentrate on the issue any given PR is doing without introducing tangential changes.

scalanet/ut/src/io/iohk/scalanet/codec/FramingCodecSpec.scala
aakoshh: Gotcha, thanks.

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
dmitry-worker: Yes, unfortunately, all asScala and asJava methods were moved to another package, from
`scala.collection.JavaConverters`  to `scala.jdk.CollectionConverters`
That's the reason I've messed up the code with java-ish stuff where the java collection is needed.

An alternative solution could be
1) Suppress warnings (deprecation)
2) Remove fatal-warnings compiler directive
3) Split the sources (2.12/src and 2.13/src) 

And I don't see anything good about those options :(

build.sc
aakoshh: I see we have a `master` branch as well. Would it be enough to merge `develop` into `master`, update the version there, then tag it, and add a pipeline step that publishes with `--release` if the tag looks like a semver? As it's described in the [Gitflow workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) I'd only use release branches as a short lived branch where any remaining bugs are sorted out before it's merged into `master` and `develop`, then get rid of them. That would avoid changes lingering there forever, not making it into the next release, etc, and also any ambiguity of whether a specific version should be looked for as a branch or as a tag.

aakoshh: I also used to have `build.sbt` take the `version` from the git tag, if it looked like a semver, so we don't have to remove the `SNAPSHOT` and then add it back when the `master` or the `release` is merged back into `develop`.

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
aakoshh: Hm, thanks for expaining. Perhaps [suppression](https://github.com/ghik/silencer#message-pattern) would be better than adding code in an ad-hoc way. @KonradStaniec wdyt?

aakoshh: https://github.com/scala/scala-collection-compat/issues/208

scalanet/discovery/src/io/iohk/scalanet/discovery/util/package.scala
aakoshh: Oh I see they renamed the method :-/ 
I wonder if a [trick like this](https://github.com/scala/scala-collection-compat/issues/208#issuecomment-497735669) works in this case as well.

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
aakoshh: Perhaps the easiest option is to do as this [comment](https://github.com/scala/scala-collection-compat/issues/208#issuecomment-497710883) suggests and only use fatal warnings as part of the 2.12 options. As long as we're compiling to 2.12, that would be our primary target, no?

scalanet/discovery/src/io/iohk/scalanet/discovery/util/package.scala
aakoshh: Or actually it should work since it's implicit.

aakoshh: This works:
```scala
  import Ordering.Implicits._

  implicit val byteOrdering: Ordering[ByteVector] =
    Ordering.by[ByteVector, IndexedSeq[Byte]](_.toIndexedSeq)
```
And the way `ByteVector.toIndexedSeq` works it doesn't create any copies either ;)

dmitry-worker: Yep, that works, I've got confused by renamed method as well :(
But anyway, it works, sorry.

scalanet/ut/src/io/iohk/scalanet/NetUtils.scala
dmitry-worker: > [scala/scala-collection-compat#208](https://github.com/scala/scala-collection-compat/issues/208)

I can do a workaround like this, but I was afraid that you wouldn't accept this monster :)

aakoshh: Well it's kinda crazy, but it would scale better. Just make sure we have a link to that explanation :smile:

aakoshh:
I wonder why we don't have any build show up in https://app.circleci.com/pipelines/github/input-output-hk/scalanet

Can't remember if it happens for draft PRs. Maybe you have to be part of some group to have your commits trigger builds?

dmitry-worker:
Closed in favor of https://github.com/input-output-hk/scalanet/pull/113

PR #113, opened 2020-12-22
[ETCM-492] Crossbuild against Scala-2.12 and Scala-2.13

dmitry-worker:


aakoshh (approved):
:beers:

aakoshh:
We just need to understand why CI isn't building your commits. @KonradStaniec do you know?

aakoshh:
And it would be good to include the ticket in the commit messages as well so we get a good history.

aakoshh:
And I forgot but could you update the `README` with examples of calling the stuff with the `csm[_]` and specific versions as well? And update the `single` example.

KonradStaniec:
@aakoshh about CI imo it has something to do that the source branch is from forked repo `dmitry-worker:ETCM-492-crossbuild-scala-2.13`. I saw this several times that CI only properly handles pr's from the same repo.

KonradStaniec:
and about your earlier comments about release cycle, I am all for updating it to some form of gitflow, with a little bit of automation. It is probably time to think about such things, as library is now used in two projects, so release cycle and some form of backward compatibility will more important now

dmitry-worker:
Need to re-create PR in order to CI procedures (see #114)

PR #114, opened 2020-12-23
[ETCM-492] - made a cross-build for scala 2.12 and 2.13

dmitry-worker:
This is a replacement for #112 and #113


jmendiola222 (approved):
Seems like this code was already approved, green light to merge if that's the case.

PR #116, opened 2021-01-05
[PM-2632] socks5 proxy support

KonradStaniec:
Easiest way to integrate with one other project is through their socks5 proxy. So to ease up initial experiments, adds socks5 proxy support to scalanet (thorough nett)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -101,7 +101,7 @@ class ScalanetModule(crossVersion: String) extends Module {
   trait ScalanetPublishModule extends PublishModule {
     def description: String
 
-    override def publishVersion = "0.5.1-SNAPSHOT"
+    override def publishVersion = "0.5.2-SNAPSHOT"
 
     override def pomSettings = PomSettings(
       description = description,
--- a/build.sc
+++ b/build.sc
@@ -12,8 +12,8 @@ class ScalanetModule(crossVersion: String) extends Module {
 
   // A cross build is now a `root` of the module:
   // > mill show csm[2.13.4].scalanet.sources
-  // [1/1] show 
-  // [1/1] show > [1/1] csm[2.13.4].scalanet.sources 
+  // [1/1] show
+  // [1/1] show > [1/1] csm[2.13.4].scalanet.sources
   // [
   //     "ref:c984eca8: ../csm/2.13.4/scalanet/src"
   // ]
@@ -24,7 +24,6 @@ class ScalanetModule(crossVersion: String) extends Module {
   //            // some output omitted //
   //     "ref:c984eca8: ../scalanet/src"
   override def millSourcePath = super.millSourcePath / os.up / os.up
-
 
   // scala 2.12 - only options
   // it causes errors when built against 2.13 (and vice-versa):
@@ -37,11 +36,10 @@ class ScalanetModule(crossVersion: String) extends Module {
     "-Ywarn-value-discard"
   )
 
-
   // scala 2.13 might have another set of options
   private val `scala 2.13 options`: Seq[String] = Nil
 
-  trait ScalanetModule extends ScalaModule { 
+  trait ScalanetModule extends ScalaModule {
     override def scalaVersion = crossVersion
 
     override def repositories =
@@ -63,7 +61,7 @@ class ScalanetModule(crossVersion: String) extends Module {
       "utf-8"
     )
 
-    private val versionOptions = crossVersion.take(4) match { 
+    private val versionOptions = crossVersion.take(4) match {
       case "2.12" => `scala 2.12 options`
       case "2.13" => `scala 2.13 options`
     }
@@ -84,7 +82,8 @@ class ScalanetModule(crossVersion: String) extends Module {
         ivy"org.scalacheck::scalacheck:1.14.0",
         ivy"ch.qos.logback:logback-core:1.2.3",
         ivy"ch.qos.logback:logback-classic:1.2.3",
-        ivy"org.mockito:mockito-core:2.21.0"
+        ivy"org.mockito:mockito-core:2.21.0",
+        ivy"com.github.mike10004:fengyouchao-sockslib:1.0.6"
       )
 
       override def moduleDeps: Seq[JavaModule] =
@@ -95,7 +94,6 @@ class ScalanetModule(crossVersion: String) extends Module {
       }
     }
   }
-
 
   // In objects inheriting this trait, use `override def moduleDeps: Seq[PublishModule]`
   // to point at other modules that also get published. In other cases such as tests
--- a/build.sc
+++ b/build.sc
@@ -101,7 +101,7 @@ class ScalanetModule(crossVersion: String) extends Module {
   trait ScalanetPublishModule extends PublishModule {
     def description: String
 
-    override def publishVersion = "0.5.2-SNAPSHOT"
+    override def publishVersion = "0.6.0-SNAPSHOT"
 
     override def pomSettings = PomSettings(
       description = description,
--- a/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
+++ b/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
@@ -190,11 +190,13 @@ object DiscoveryNetwork {
       private def maybeRespond[Res](maybeResponse: Task[Option[Res]])(
           f: Res => Task[Unit]
       ): Task[Unit] =
-        maybeResponse.onErrorRecoverWith {
-          case NonFatal(ex) =>
-            // Not responding to this one, but it shouldn't stop handling further requests.
-            Task(logger.error(s"Error handling incoming request: $ex")).as(None)
-        }.flatMap(_.fold(Task.unit)(f))
+        maybeResponse
+          .onErrorRecoverWith {
+            case NonFatal(ex) =>
+              // Not responding to this one, but it shouldn't stop handling further requests.
+              Task(logger.error(s"Error handling incoming request: $ex")).as(None)
+          }
+          .flatMap(_.fold(Task.unit)(f))
 
       /** Serialize the payload to binary and sign the packet. */
       private def pack(payload: Payload): Task[Packet] =
@@ -382,9 +384,11 @@ object DiscoveryNetwork {
                 val next = (acc: Z) => Some(acc -> (count + 1))
                 Task.pure(f(acc, response).bimap(next, next))
 
-              case (None, _) => 
+              case (None, _) =>
                 // Invalid state - this cannot happen
-                Task.raiseError(new IllegalStateException(s"Unexpected state while collecting responses from ${channel.to}"))
+                Task.raiseError(
+                  new IllegalStateException(s"Unexpected state while collecting responses from ${channel.to}")
+                )
             }
             .map(_.map(_._1))
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
@@ -2,10 +2,13 @@ package io.iohk.scalanet.peergroup
 
 import cats.effect.Resource
 import io.iohk.scalanet.peergroup.Channel.ChannelEvent
+import io.iohk.scalanet.peergroup.PeerGroup.PeerGroupWithProxySupport.Socks5Config
 import io.iohk.scalanet.peergroup.PeerGroup.ServerEvent
 import monix.eval.Task
 import monix.reactive.Observable
 import scodec.Codec
+
+import java.net.InetSocketAddress
 
 /**
   * A Channel represents a route between two peers on the network for the purposes
@@ -106,6 +109,30 @@ trait PeerGroup[A, M] {
 object PeerGroup {
 
   abstract class TerminalPeerGroup[A, M](implicit codec: Codec[M]) extends PeerGroup[A, M]
+
+  abstract class PeerGroupWithProxySupport[A, M](implicit codec: Codec[M]) extends PeerGroup[A, M] {
+
+    /**
+      * This method builds a communication channel for the current peer to communicate messages with the
+      * desired address, going through proxy specified in proxy config
+      *
+      * @param to the address of the entity that would receive our messages. Note that this address can
+      *           be the address to refer to a single peer as well as a multicast address (to refer to a
+      *           set of peers).
+      * @param proxyConfig address and optional auth options to socks5 proxy
+      * @return the channel to interact with the desired peer(s).
+      */
+    def client(to: A, proxyConfig: Socks5Config): Resource[Task, Channel[A, M]]
+  }
+
+  object PeerGroupWithProxySupport {
+    final case class Sock5AuthenticationConfig(user: String, password: String)
+
+    final case class Socks5Config(
+        proxyAddress: InetSocketAddress,
+        authConfig: Option[Sock5AuthenticationConfig]
+    )
+  }
 
   sealed trait ServerEvent[A, M]
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
@@ -126,11 +126,11 @@ object PeerGroup {
   }
 
   object ProxySupport {
-    final case class Sock5AuthenticationConfig(user: String, password: String)
+    final case class Socks5AuthenticationConfig(user: String, password: String)
 
     final case class Socks5Config(
         proxyAddress: InetSocketAddress,
-        authConfig: Option[Sock5AuthenticationConfig]
+        authConfig: Option[Socks5AuthenticationConfig]
     )
 
     def apply[A, M](config: Socks5Config)(underlying: PeerGroup[A, M] with ProxySupport[A, M]): PeerGroup[A, M] =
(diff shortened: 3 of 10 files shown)


aakoshh:
The ticket didn't have any description, could you elaborate a little bit in the PR on what this is about? For example is it a CLI app that starts a proxy like `ssh` that you use from your browser? How is one supposed to use it?

build.sc
```diff
@@ -96,14 +95,13 @@ class ScalanetModule(crossVersion: String) extends Module {
...
 
-
   // In objects inheriting this trait, use `override def moduleDeps: Seq[PublishModule]`
   // to point at other modules that also get published. In other cases such as tests
   // it can be `override def moduleDeps: Seq[JavaModule]`.
   trait ScalanetPublishModule extends PublishModule {
     def description: String
 
-    override def publishVersion = "0.5.1-SNAPSHOT"
+    override def publishVersion = "0.5.2-SNAPSHOT"
```
aakoshh: If this is a new feature and not a bugfix, shouldn't the version should be `0.6.0-SNAPSHOT`?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala
```diff
@@ -382,9 +384,11 @@ object DiscoveryNetwork {
                 val next = (acc: Z) => Some(acc -> (count + 1))
                 Task.pure(f(acc, response).bimap(next, next))
 
-              case (None, _) =>
```
aakoshh, line 385: @dmitry-worker you may want to configure your editor to trim trailing spaces.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -107,6 +110,30 @@ object PeerGroup {
...
+      * This method builds a communication channel for the current peer to communicate messages with the
+      * desired address, going through proxy specified in proxy config
+      *
+      * @param to the address of the entity that would receive our messages. Note that this address can
+      *           be the address to refer to a single peer as well as a multicast address (to refer to a
+      *           set of peers).
+      * @param proxyConfig address and optional auth options to socks5 proxy
+      * @return the channel to interact with the desired peer(s).
+      */
+    def client(to: A, proxyConfig: Socks5Config): Resource[Task, Channel[A, M]]
```
aakoshh, line 125: Having the config in the `client` method is a bit unexpected. If I set a SOCKS5 proxy in my browser I expect it to transparently work for all the communication. This suggests that the calling code should actively know that they have to establish the connection through a proxy, and a different proxy or credentials may be required for every client. 

I'd expected that the the `PeerGroup` implementation itself would receive the `proxyConfig` in its constructor and handle the proxying in a way that `client(to = remote)` keeps working transparently, so that the decision to use a proxy is left in the top level composition root.

If the proxy varies per remote address, couldn't the the config be deducted from the `to` parameter itself, passed to the constructor of the implementation, to keep this interface as it was?

Since the ticket has no description, what exactly is the use case we have to support?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -115,6 +118,16 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
           val pipeline = ch.pipeline()
           val sslHandler = sslClientCtx.newHandler(ch.alloc())
 
+          socks5Config.foreach { config =>
+            val sock5Proxy = if (config.authConfig.isDefined) {
+              val authConfig = config.authConfig.get
+              new Socks5ProxyHandler(config.proxyAddress, authConfig.user, authConfig.password)
+            } else {
+              new Socks5ProxyHandler(config.proxyAddress)
+            }
```
aakoshh:
```suggestion
            val sock5Proxy = config.authConfig.fold {
               new Socks5ProxyHandler(config.proxyAddress)
            } { authConfig =>
              new Socks5ProxyHandler(config.proxyAddress, authConfig.user, authConfig.password)
            } 
```
Just to avoid the `.get`.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -60,6 +65,33 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+              Observable(ch).zip(ch.channelEventObservable.guarantee(release))
+
+          }.headL
+          sessions <- Task(proxy.getSessionManager.getManagedSessions)
+          (chan, rec) = wo
+        } yield {
+          assert(sessions.size() == 1)
+          val session = sessions.get(1.toLong)
+          assert(session.getNetworkMonitor.getReceiveTCP > 0)
+          assert(session.getNetworkMonitor.getSendTCP > 0)
```
aakoshh, line 96: Is `getReceiveTCP` tracking what `ch1` sent to the proxy, and `getSendTCP` what the proxy sent  to the `server`? Is there a separate count for responses coming from the `server`?

aakoshh (commented):
Nice addition, viva la Netty! 

But I maintain that the proxy config should be passed to the `DynamicTLSPeerGroup` and not the `client` method, as this looks more like a deployment concern that you'd want to be able to switch on/off via configuration, rather than sprinkle around your codebase, where it's not even optional: you either call `.client(to)` and there's no proxy support or `.client(to, socksConfig)` and there _has_ to be one, and not for example `.client(to, maybeSocksConfig)`, although that would be inconvenient too, if in most cases it's not required.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
KonradStaniec: thats my guess, but the library does not have great docs so it is as good as yours. From what i known that all are stats for tcp we can get. There is also the same pair of stats for udp.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
KonradStaniec: So i have added more context to the ticket https://jira.iohk.io/browse/PM-2632 ( sorry for not doing this in the first place).

Initially have also thought that passing that info in config would be optimal approach, but after some thought I came to conclusion that that there are few arguments for having it per client basis:

1. Traffic going through proxy will have lower latency and throughput but be private. There is possibility that user of PeerGroup (our node) will want to connect to trusted nodes without using the proxy (as some so sort of allow-list).
2. If the network by which proxy route traffic is down for some reasons, then it easy for the user of PeerGroup (our node) to detect that, and try to do something like: try to connect 3 times through proxy, if it fails, connect without it to check if it problem with network or proxied network.
3. this is experimental feature and having it in client is just more flexible and will ease up experiments (minor argument)

KonradStaniec: as to second question, proxy not vary per remote connection so passing this info in `to` param is not something i would do (as proxy in not part of remote node identity but actually a part of local node deployment)

aakoshh: Thanks for the extra context about using Nym!

The points you made still make me think that hiding this inside the peer group, without surfacing it in the `client` method, would be preferrable.

> There is possibility that user of PeerGroup (our node) will want to connect to trusted nodes without using the proxy (as some so sort of allow-list).

First of all, sorry, I was vague with my second question, I didn't want to suggest that the parameters are put _in_ the `to` but rather that a function is passed to the constructor that can decide if a given `to` needs to be proxied:
```scala
new DynamicTLSPeerGroup[A](
  ...,
  addressSocksConfig: A => Option[Socks5Config] = 
    a => if (allowList.contains(a)) then None else Some(socksConf)
)
```
The benefit is that this allow-list wouldn't have to be pushed everywhere and you can add it at a later stage without going through all the code that just uses the peer group. It looks like an optimisation, not something that is a core part of program flow. 

> If the network by which proxy route traffic is down for some reasons, then it easy for the user of PeerGroup (our node) to detect that, and try to do something like: try to connect 3 times through proxy, if it fails, connect without it to check if it problem with network or proxied network.

I'd think it's even easier for the peer group itself to detect that, since it's the one that handles all the connection logic. As a user I'd rather like if I could just do `.client(to = remote)` and the peer group would pick whichever alternative proxy, or direct connection works, rather than me having to do such logic in my code. 

> this is experimental feature and having it in client is just more flexible and will ease up experiments (minor argument)

That is also an argument for not having to modify existing code using the peer groups, but just enable it behind the scenes, and if it doesn't work then disable it in config.

KonradStaniec: I think the core issue here is how we view `DynamicTLSPeerGroup`. In my opinion main responsibility of it is to connect to other peers and receive connections (channels) from other peers. I view having explicit function saying `connect to peer X but through proxy` as something in line with this responsibilities, but imo `allow-list` or any reconnect strategies are the responsibility of higher layer protocol. (lets call it `OurNodePeerGroup`, it still does not existis but should in the future)

> That is also an argument for not having to modify existing code using the peer groups, but just enable it behind the scenes, and if it doesn't work then disable it in config

I do not quite follow this one, user of peer group can have such config, and explicitly chose either normal connection or connection by proxy himself

aakoshh: > I do not quite follow this one

What I'm trying to say is that the _code site using_ the peer group shouldn't have to have the responsibility of choosing between normal and proxied connections. For example the `DiscoveryNetwork` has a `PeerGroup` but it doesn't know that it's a static or dynamic UDP peer group or a TCP one; similarly it shouldn't have to worry about connecting through a proxy. Instead, the _human user_ configuring the system should make sure that they are passing in a peer group which is correctly configured to either use a proxy or not, or maybe not use one for LAN connections but use one for outgoing ones, etc. 

Because this is, as you said, an experimental feature, it's much easier to experiment with it by passing in a modified peer group that transparently handles this to _any_ component that uses Scalanet, than to go and modify all of them to start passing around proxy configs and decide whether they should connect directly or through proxy on a case by case basis.

aakoshh: Okay I see what you are saying, that we'll use these through a peer group like the `OurNodePeerGroup`, not directly.

It would be perhaps good to make the socks supporting method internal to Scalanet while it's experimental andnalso to add a basic wrapper method to the `PeerGroupWithSocksSupport` companion that takes one with support and turns it into a regular one that adds to proxy config to the calls. It would indicate the suggested use of this functionality, before people start putting the logic all over the place.

aakoshh: Although I guess making it internal would make experimenting tedious with our release cycle, but it would mean the benefits can be shared by all and not be locked away in another project.

KonradStaniec: so going in order:

>  For example the DiscoveryNetwork has a PeerGroup but it doesn't know that it's a static or dynamic UDP peer group or a TCP one

Imo it should, as current kademlia makes little sense over lets say current tls, but this probably other issue that we lack abstraction which would determine transport,as some things are suited for udp or tcp.

>  Instead, the human user configuring the system should make sure that they are passing in a peer group which is correctly configured to either use a proxy or not, or maybe not use one for LAN connections but use one for outgoing ones, etc.

I slightly disagree here, it is up to the caller/user of the library to determine if he wants to use proxied connection or not. Maybe he wants to use the proxy only for one particular connection and not for others, and we won't be able to express all things in config at this level. And if we care about end user configuring the system, we can have such configuration on our node layer, and use primitives which library provides.

> Because this is, as you said, an experimental feature, it's much easier to experiment with it by passing in a modified peer group that transparently handles this to any component that uses Scalanet, than to go and modify all of them to start passing around proxy configs and decide whether they should connect directly or through proxy on a case by case basis.

Imo this a little exaggeration, user of library can create one config and pass it everywhere also `.client` is also called rarely (usually one per-peer) and all operations are on created channel. Also it is up to that call site to decide if the connection should be proxied or not. Or how error should be handled is allso callstie concern, either decalre failure or try to reconnect without proxy. 

So summarising, imo at library level it is better to expose explicitly possibly of creating proxied or not-proxied connections. The config for node user can be exposed at higher layer (our node) and can have all the options mentioned earlier.

KonradStaniec: > It would be perhaps good to make the socks supporting method internal to Scalanet while it's experimental andnalso to add a basic wrapper method to the PeerGroupWithSocksSupport companion that takes one with support and turns it into a regular one that adds to proxy config to the calls. It would indicate the suggested use of this functionality, before people start putting the logic all over the place.

Could you show code example how it wold like ? ( I am not sure from description)

aakoshh: Of course, sorry I was typing on my phone. 

> user of library can create one config and pass it everywhere

Not really because they either call `client.to(remote)` or `client.to(remote, proxyConfig)`, the version that takes an optional config is not exposed.  

Anyway, here's how I thought you could nudge people into using this functionality without mixing the decisions you talk about into their end-user code:

```scala
trait ProxySupport[A, M] {
   protected[scalanet] def client(to: A, proxyConfig: Socks5Config): Resource[Task, Channel[A, M]]
}

object ProxySupport {
  final case class Sock5AuthenticationConfig(...)
  final case class Socks5Config(...)
  
  /** Wrap a peer group which supports proxying into a peer group that transparently and unconditionally proxies connections. */
  def apply[A, M](config: Socks5Config)(underlying: PeerGroup[A, M] with ProxySupport[A, M]): PeerGroup[A, M] = 
    new PeerGroup[A, M] {
       override def processAddress = underlying.processAddresss
       override def client(to: A) = underlying.client(to, config)
       override def nextServerEvent = underling.nextServerEvent
    }
}
```

And then the test would look like: 
```scala
 (for {
      client <- ProxySupport(Socks5Config(new InetSocketAddress("localhost", 1080), None) { 
        DynamicTLSPeerGroup[String](getCorrectConfig())
      }
      proxy <- buildProxyServer(1080)
      server <- DynamicTLSPeerGroup[String](getCorrectConfig())
      ch1 <- client.client(server.processAddress)
    } yield (server, ch1, proxy)).use {
```

Functions like allow-list and the option to try different proxies if there are failures would be implemented in similar middlewares, but end users code would always deal with `PeerGroup`.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -107,6 +110,30 @@ object PeerGroup {
 
   abstract class TerminalPeerGroup[A, M](implicit codec: Codec[M]) extends PeerGroup[A, M]
 
+  abstract class PeerGroupWithProxySupport[A, M](implicit codec: Codec[M]) extends PeerGroup[A, M] {
```
aakoshh: By the way I'm not sure what the significance of `TerminalPeerGroup` was but this could (as in the example I posted below) stay a trait, since the `DynamicTLSPeerGroup` implementing this feature is still a terminal one, while as I've shown this could be done with a non-terminal wrapper group.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
KonradStaniec: I really like it! 

Do you think it makes sense to have this and to have `def client(to: A, proxyConfig: Socks5Config): Resource[Task, Channel[A, M]]` public ?  It would be good indication of principled usage but on the other hand we will be able to quickly experiment on our node side.

aakoshh: Cool! 

That's what I'm wondering myself too. Making it internal to scalanet forces people to contribute to the project but makes the feature look a bit crippled because you can't access it to make your own wrappers in your project. A simple `protected` is not much use because we create the terminal groups via one of the `apply` methods that return a `Resource` of a concrete class, so it's not as easy to inherit a base class to get access to it.

So yeah, `public` is more useful for experimentation. I'd leave it up to your judgement.

aakoshh: I used `mill __.publishLocal` for local experiments.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -107,6 +110,30 @@ object PeerGroup {
...
+      *           be the address to refer to a single peer as well as a multicast address (to refer to a
+      *           set of peers).
+      * @param proxyConfig address and optional auth options to socks5 proxy
+      * @return the channel to interact with the desired peer(s).
+      */
+    def client(to: A, proxyConfig: Socks5Config): Resource[Task, Channel[A, M]]
+  }
+
+  object PeerGroupWithProxySupport {
+    final case class Sock5AuthenticationConfig(user: String, password: String)
```
aakoshh:
```suggestion
    final case class Socks5AuthenticationConfig(user: String, password: String)
```

aakoshh (approved):
Great, thanks for indulging me in the discussions, we had similar principles all along :+1:

KonradStaniec:
thanks to you too :) I found that such discussions usually make code much better

PR #119, opened 2021-01-12
[PM-2653] Add automation to release.

KonradStaniec:
- Add developers to build (--release flag requires it)
- add release to maven when merging to master


.circleci/config.yml
```diff
@@ -47,9 +45,20 @@ jobs:
...
+                  - run:
+                      name: publish into Sonatype
+                      command: .circleci/publish
+            - when:
+                condition:
+                  equal: [ master, << pipeline.git.branch >> ]
+                steps:
+                  - run:
+                      name: publish release
+                      command: .circleci/publishRelease
```
aakoshh, line 51: It looks like we're building up quite a bit of redundancy: both `publish` and `publishRelease` have checks inside them to see they are on the right branch, and these have to be matched here by the conditional checks too. 

Wouldn't it be easier at this point to have just a single publish script that either:
* calls with or without `--release true` depending on the branch, and then no conditional check is needed here, or
* keep the conditional and call `publish` or `publish --release true` depending on which one matched, passing on the extra arguments with `$@` inside?

build.sc
```diff
@@ -109,7 +109,17 @@ class ScalanetModule(crossVersion: String) extends Module {
...
+      developers = Seq(
+        Developer("aakoshh", "Akosh Farkash", "https://github.com/aakoshh"),
+        Developer("shaileshp0110", "Shailesh Patil", "https://github.com/shaileshp0110"),
+        Developer("jtownson", "Jeremy Townson", "https://github.com/jtownson"),
+        Developer("KonradStaniec", "Konrad Staniec", "https://github.com/KonradStaniec"),
+        Developer("marcesquerra", "Marc Esquerra", "https://github.com/marcesquerra"),
+        Developer("EzequielPostan", "Ezequiel Postan", "https://github.com/EzequielPostan"),
+        Developer("AlexITC", "Alexis Hernandez", "https://github.com/AlexITC"),
+        Developer("dmitry-worker", "Dmitry Voronov", "https://github.com/dmitry-worker"),
+        Developer("krcz", "Marcin Kurczych", "https://github.com/krcz")
```
aakoshh:
```suggestion
        Developer("aakoshh", "Akosh Farkash", "https://github.com/aakoshh"),
        Developer("AlexITC", "Alexis Hernandez", "https://github.com/AlexITC"),
        Developer("dmitry-worker", "Dmitry Voronov", "https://github.com/dmitry-worker"),
        Developer("EzequielPostan", "Ezequiel Postan", "https://github.com/EzequielPostan"),
        Developer("jtownson", "Jeremy Townson", "https://github.com/jtownson"),
        Developer("KonradStaniec", "Konrad Staniec", "https://github.com/KonradStaniec"),
        Developer("krcz", "Marcin Kurczych", "https://github.com/krcz"),
        Developer("marcesquerra", "Marc Esquerra", "https://github.com/marcesquerra"),
        Developer("shaileshp0110", "Shailesh Patil", "https://github.com/shaileshp0110")
```
Alphabetical order? I always win :)

aakoshh (commented):
At what step are we supposed to remove `-SNAPSHOT` from the version before release? I thought maybe we could allow some extra commits to go into master, and only publish if it's tagged?

.circleci/config.yml
KonradStaniec: Maybe you are right here. I will got with `1` approach as i do not like `$@` as i am pretty sure everyone would need to always check what it means

aakoshh:
I'm just looking at the [mill docs](http://www.lihaoyi.com/mill/#deploying-your-code) and it says `--release` is `true` by default :thinking: 

The version on [github](https://github.com/lihaoyi/mill/blob/master/docs/pages/3%20-%20Common%20Project%20Layouts.md#publishing) seems to be more up to date.

KonradStaniec:
So my plan about flow and removing snapshot was something about that:

1. Create release branch , remove snapshot here, make pr to master, remove branch after merge, tag merge commit
2. Create merge back pr from master do develop, attach new snapshot here, merge to develop

aakoshh:
Look mill has a [Version](https://github.com/lihaoyi/mill/blob/d6dd9f619984335b23a3c3e4fa4eed9fa77ade29/contrib/versionfile/src/mill/contrib/versionfile/Version.scala) which automatically attaches and removes the `-SNAPSHOT` depending on whether it's a release, but I'm not sure if it's internal.

KonradStaniec:
Hmm you are talking about something like that - http://www.lihaoyi.com/mill/page/contrib-modules.html#version-file ? 

Tbh I would do it in separate pr, as probably geting to proper flow with it would require a little experimentation.

aakoshh:
Yes, you're right it's part of `version-file`. I see it even has git commands to tag and push, however there could be a very minimalistic approach:

Change `build.sc` to take the version from the file:
```scala
def publishVersion = versionFile.currentVersion().toString
```
Add the current version on `develop` and commit:
```console
$ 0.6.0-SNAPSHOT > versionFile/version
$ git add .
$ git commit
```

Whenever there's a new feature or patch we bump the appropriate version on the release branch or feature PR but leave the `-SNAPSHOT` in tact.
```console
$ mill versionFile.setNextVersion --bump minor
$ git add .
$ git commit
```

In the `publish` script if it's a `--release` then we remove `-SNAPSHOT` without committing it.
```sh
if [[ "$CIRCLE_BRANCH" == "master" ]]; then
  mill versionFile.setReleaseVersion
  mill mill.scalalib.PublishModule/publishAll \
     ...
fi
```

This way there's no reason to add and remove it manually.

aakoshh:
In any case it would be good to add the workflow description to the README. Maybe halt publishing on `master` if it's `-SNAPSHOT`?

KonradStaniec:
I like this minimalistic approach! Lets go with this way then. 
One change i would make it is to bump minor version when merging back from master to develop, that way the version on develop would be unreleased i.e
Now we are working on `0.6.0-SNAPSHOT` so the flow to release would be:
1. create release branch `release/0.6.0` targeting `master`
2. merge it to master, tag it. (now we have released version `0.6.0`
3. create merge back pr from master to develop bumping minor version to `0.7.0-SNAPSHOT`
wdyt ?

aakoshh:
> create merge back pr from master to develop bumping minor version to 0.7.0-SNAPSHOT

You are right, the next change on `develop` should be a `--bump minor`, i.e. a new feature, because if we release `0.6.0` then any fixes to that should happen as hotfixes on the `master` with `--bump patch`, e.g. `0.6.1`. That should be merged back into `develop`. We shouldn't push more updates to the same snapshot from `develop`.

Because we would not be committing the version during the release, technically there may not be any change on `master` to merge back into `develop`, but I guess a formal PR is a good way to signal that this happened and prevent accidental overwrites.

aakoshh:
I had another idea once that would completely automate the version updates: instead of manually bumping the version, a PR would be required to append a line to the `CHANGELOG.md` and indicate whether it's a `major`, `minor` or `patch` thing, together with the description. When it comes to release, the process would automatically detect the changes since the last one and increment the appropriate parts of the version string (whether once or multiple times I don't know) and commit the changes back. 

This way any further change to `develop` would be an automatic increase in something. And merging the changelog after a hotfix from `master` to `develop` would let the developer arrange the order, so in the end if there's a new feature `0.7.0-SNAPSHOT` on `develop` and a hotfix `0.6.1` on `master` then the hotfix could be merged into the changelog _before_ the feature, which results in `0.7.0-SNAPSHOT` being unchanged. Or maybe it should go on top of it and become `0.7.1-SNAPSHOT`.

README.md
```diff
@@ -99,6 +99,13 @@ To use it in a downstream project add the snapshots to the resolvers, e.g. in `b
 resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"
 ```
 
+To publish new release to maven central:
+1. Create release branch `realease/versionFromVersionFile`
+2. Crete pr from release branch to `master` branch
```
aakoshh, line 104:
```suggestion
1. Create release branch `realease/<versionFromVersionFile>`
2. Create PR from release branch to `master` branch
```

README.md
```diff
@@ -99,6 +99,13 @@ To use it in a downstream project add the snapshots to the resolvers, e.g. in `b
...
+3. Merge release pr to `master`, tag merge commit and push it
+4. Create merge back pr from `master` to `develop` bumping appropriate version in `versionFile/version`
```
aakoshh:
```suggestion
1. Create release branch `realease/versionFromVersionFile`
2. Create PR from release branch to `master` branch
3. Merge release PR to `master`, tag merge commit and push it
4. Create merge back PR from `master` to `develop` bumping appropriate version in `versionFile/version`, e.g. `mill versionFile.setNextVersion --bump minor`, to make sure no more updates are sent to the released snapshot
```

aakoshh:
One thing I'm not certain about in your workflow is hotfixes:
1. Release `0.6.0`
2. Merge `master` into `develop` with a bump. Now `develop` is `0.7.0-SNAPSHOT` and `master` is `0.6.0-SNAPSHOT`. 
3. Apply a hotfix on `master`, bumping it to `0.6.1-SNAPSHOT`, released as `0.6.1`. 
4. Merge `master` into `develop`: do you bump to `0.6.2-SNAPSHOT`? Or `0.7.0-SNAPSHOT` again? Or `0.7.1-SNAPSHOT` since it now has a fix it didn't have yesterday? Definitely not `0.8.0-SNAPSHOT`. No point in `0.6.2-SNAPSHOT` either because on `develop` we'll release features, not fixes.

aakoshh (approved):
Cool, it looks like we'll be able to manage this in PRs fairly easily!

KonradStaniec:
Yep hot fixes are quite tricky. My thinking was to, create hotfix branch and:

- either target both master and develop ( but do not change version in develop)
- or target master and then merge it back to develop without changing develop versions (in your example leave `0.7.0-SNAPSHOT`)

My thinking is that, by having this automation, we can release more often and do not depend on snapshot versions as much. So we will not be required to bump versions with each pr, but have normal release cycle, and as downstream projects will not depend on snapshots as much we will not break ale the nix checks.

PR #121, opened 2021-01-14
Add larger timeouts in release script

KonradStaniec:
It seems timeouts for release should be higher - values copied from https://github.com/lihaoyi/mill/blob/master/ci/release-maven.sh


aakoshh (approved):
You found that pretty quick!

PR #123, opened 2021-01-15
[Fix] Increase release timeout on CI

KonradStaniec:
During last release `0.6.0` CircleCi failed halfway during publishing process, this led to release being not finished. I needed to close and release staging repository manually. 

This pr increases this timeout to half and hour which should be enough.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,7 @@ jobs:
             - run:
                 name: publish
                 command: .circleci/publish
+                no_output_timeout: 30m
 
       # scoverage disabled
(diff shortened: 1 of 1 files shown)


aakoshh (approved):
So the timeouts in the `publish` script are ~1 minute~, but for CI it's 30? It should see some positive or negative output after a minute, shouldn't it?

KonradStaniec:
So time outs are set to `600000` which i understand is `10min`, and this is set only per one http read. And as I understand this is per read timeout, and there can be multiple of those during whole release process.

PR #124, opened 2021-02-02
[PM-2726] Add new validations from libp2p specs

KonradStaniec:
# Description

Add new validations from libp2p specs: 
- certificate self-signature validation
- validation that certificate signature is correct type - we force all signatures in dynamic tls peergroup to be ecdsa with sha256
- validation that certificate use NamedCurve encoding for elliptic curve parameters (public key)
- validation that certificate does not contain unknown critical extensions - in our case it should have at most one extension, which is extension to encode host public key

Also created ticket to refactor a little crypto code - https://jira.iohk.io/browse/PM-2744 as with all those additions it is getting a little out of hand.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
@@ -5,7 +5,6 @@ import java.security._
 import java.security.cert.X509Certificate
 import java.security.spec.{ECGenParameterSpec, PKCS8EncodedKeySpec, X509EncodedKeySpec}
 import java.util.Date
-
 import io.iohk.scalanet.peergroup.dynamictls.DynamicTLSExtension.Extension
 import org.bouncycastle.asn1.sec.SECNamedCurves
 import org.bouncycastle.asn1.x500.X500Name
@@ -39,7 +38,8 @@ private[scalanet] object CryptoUtils {
 
   type SignatureBytes = Array[Byte]
 
-  private val usedSignatureScheme = "SHA256withECDSA"
+  abstract class SignatureScheme(val name: String)
+  case object SHA256withECDSA extends SignatureScheme("SHA256withECDSA")
 
   private val usedKeyScheme = "EC"
 
@@ -51,7 +51,7 @@ private[scalanet] object CryptoUtils {
     new ECDomainParameters(curveParams.getCurve, curveParams.getG, curveParams.getN, curveParams.getH)
 
   private def getEcdsaSignature: Signature = {
-    Signature.getInstance(usedSignatureScheme, PROVIDER)
+    Signature.getInstance(SHA256withECDSA.name, PROVIDER)
   }
 
   private def getEcKeyFactory: KeyFactory = {
@@ -99,7 +99,7 @@ private[scalanet] object CryptoUtils {
     new KeyPair(keyFac.generatePublic(spkiKeySpec), keyFac.generatePrivate(pkcs8KeySpec))
   }
 
-  def getKeyFromBytes(bytes: Array[Byte]): Try[PublicKey] = Try {
+  def getSecp256k1KeyFromBytes(bytes: Array[Byte]): Try[PublicKey] = Try {
     val ecPoint = curve.getCurve.decodePoint(bytes)
     val spec = new ECParameterSpec(curveParams.getCurve, curveParams.getG, curveParams.getN)
     val pubKeySpec = new ECPublicKeySpec(ecPoint, spec)
@@ -122,7 +122,8 @@ private[scalanet] object CryptoUtils {
       random: SecureRandom,
       extensions: List[Extension],
       beforeDate: Date,
-      afterDate: Date
+      afterDate: Date,
+      signatureScheme: SignatureScheme
   ): X509Certificate = {
     val name = "scalanet-tls"
     val sn = new BigInteger(64, random)
@@ -134,7 +135,7 @@ private[scalanet] object CryptoUtils {
       certificateBuilder.addExtension(extension.oid, extension.isCritical, extension.value)
     }
 
-    val signer = new JcaContentSignerBuilder(usedSignatureScheme).build(connectionKeyPair.getPrivate);
+    val signer = new JcaContentSignerBuilder(signatureScheme.name).build(connectionKeyPair.getPrivate);
 
     val ca = certificateBuilder.build(signer)
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
@@ -12,8 +12,6 @@ import java.security.interfaces.ECPublicKey
 import scala.util.{Failure, Success, Try}
 
 private[scalanet] object CustomTlsValidator {
-  private val knownCertificateExtension = SignedKey.extensionIdentifier
-
   sealed abstract class CertificateError(msg: String) extends CertificateException(msg)
   case object WrongNumberOfCertificates extends CertificateError("Number of certificates not equal 1")
   case object WrongCertificateDate extends CertificateError("Certificate is expired or not yet valid")
@@ -48,11 +46,7 @@ private[scalanet] object CustomTlsValidator {
   private def validateCertificatesQuantity(
       certificates: Array[X509Certificate]
   ): Either[CertificateError, X509Certificate] = {
-    if (certificates.length != 1) {
-      Left(WrongNumberOfCertificates)
-    } else {
-      Right(certificates.head)
-    }
+    Either.cond(certificates.length == 1, certificates.head, WrongNumberOfCertificates)
   }
 
   /**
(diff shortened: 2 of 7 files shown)


scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -58,22 +81,41 @@ object CustomTlsValidator {
...
+    *
+    * Taking both into account we force using "SHA256withECDSA"
+    */
+  private def validateCertificateSignatureScheme(cert: X509Certificate): Either[CertificateError, Unit] = {
+    val sigAlgName = cert.getSigAlgName
+    if (sigAlgName.equalsIgnoreCase(CryptoUtils.SHA256withECDSA.name)) {
+      Right(())
+    } else {
+      Left(WrongCertificateSignatureScheme)
+    }
```
aakoshh: I was wondering if `Either.cond` would be better, like in `validateServerId`. Same for `validateCertificatesQuantity`?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -85,22 +127,24 @@ object CustomTlsValidator {
...
-      _ <- validateCertificateSignature(signedKeyExtension, cert)
-      _ <- if (connectingTo.isDefined) validateServerId(signedKeyExtension, connectingTo.get) else Right(())
-    } yield {
-      signedKeyExtension
+  def validateCertificatePublicKey(cert: X509Certificate): Either[CertificateError, ECPublicKey] = {
+    val certificatePublicKey = cert.getPublicKey
+    Try {
+      val ecPublicKey = certificatePublicKey.asInstanceOf[ECPublicKey]
+      val subjectKeyInfo = SubjectPublicKeyInfo.getInstance(ASN1Primitive.fromByteArray(ecPublicKey.getEncoded))
+      val x962Params = X962Parameters.getInstance(subjectKeyInfo.getAlgorithm.getParameters)
```
aakoshh, line 129: How did you figure this out?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -85,22 +127,24 @@ object CustomTlsValidator {
...
+      (ecPublicKey, x962Params)
+    }.toEither.left.map(_ => WrongCertificateKeyFormat).flatMap {
+      case (key, parameters) =>
+        if (parameters.isNamedCurve) {
+          Right(key)
+        } else {
+          Left(WrongCertificateKeyFormat)
+        }
     }
```
aakoshh:
```suggestion
    } match {
      case Success((key, parameters)) if parameters.isNamedCurve =>
        Right(key)
      case _ =>
          Left(WrongCertificateKeyFormat)
    }
```
If you wanted to match the style of the other `Try` constructs.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -109,16 +153,48 @@ object CustomTlsValidator {
...
-    val extension = cert.getExtensionValue(extensionId)
-    if (extension == null) {
-      Left(NoCertExtension)
-    } else {
-      Right(extension)
+    Option(cert.getExtensionValue(extensionId)).toRight(NoCertExtension).flatMap { extension =>
+      val criticalExtensions = cert.getCriticalExtensionOIDs
+
+      val containsOnlyOneKnownCriticalExtension = criticalExtensions.size() == 0 ||
+        (criticalExtensions.size() == 1 && criticalExtensions.contains(knownCertificateExtension))
```
aakoshh:
```suggestion
        (criticalExtensions.size() == 1 && criticalExtensions.contains(SignedKey.extensionIdentifier))
```
Since there's only one `knownCertificateExtension`, I reckon it would be easier if we didn't have to scroll up to the top to see what it is, or if the variable was declared here, so we'd know it's not used for anything other than this check.

Actually since `validateCertificates` passes in `SignedKey.extensionIdentifier` in `extendsionId`, would it make sense to just use `extensionId` in the `contains` check?

Or maybe to break out this check, since it looks like `validateCertificateExtension` is about getting a specific (possibly non-critical) extension, and if we had more than one, it wouldn't make too much sense to validate all known critical extensions every time we read the value of one of the ones we need.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -109,16 +153,48 @@ object CustomTlsValidator {
...
       cert: X509Certificate,
       extensionId: String
   ): Either[CertificateError, Array[Byte]] = {
-    val extension = cert.getExtensionValue(extensionId)
-    if (extension == null) {
-      Left(NoCertExtension)
-    } else {
-      Right(extension)
+    Option(cert.getExtensionValue(extensionId)).toRight(NoCertExtension).flatMap { extension =>
+      val criticalExtensions = cert.getCriticalExtensionOIDs
```
aakoshh: Is it possible to have `getExtensionValue` return `Some` for the signed key, but then for it not to be marked critical, if we know that it's a critical extension?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
```diff
@@ -109,16 +153,48 @@ object CustomTlsValidator {
...
+      connectingTo: Option[BitVector]
+  ): Either[CertificateError, SignedKey] = {
+    for {
+      cert <- validateCertificatesQuantity(certificates)
+      _ <- validateCertificateSignatureScheme(cert)
+      _ <- validateCertificateSelfSig(cert)
+      _ <- validateCertificateDate(cert)
+      validPublicKey <- validateCertificatePublicKey(cert)
+      extension <- validateCertificateExtension(cert, SignedKey.extensionIdentifier)
+      signedKeyExtension <- validateExtension(extension)
```
aakoshh: And this `validateExtension` look like it could be dealing with any extension, but in reality it's hardcoded to `SignedKey` as well. 

So perhaps something like this?
```scala
_ <- validateOnlyKnownCriticalExtensions(cert)
signedKeyExtension <- getCertificateExtension(cert, SignedKey.extensionIdentifier).flatMap(validateSignedKeyExtension)
```

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
```diff
@@ -122,7 +122,8 @@ private[scalanet] object CryptoUtils {
       random: SecureRandom,
       extensions: List[Extension],
       beforeDate: Date,
-      afterDate: Date
+      afterDate: Date,
+      signatureScheme: SignatureScheme
```
aakoshh, line 126: If `SHA256withECDSA` is the only one allowed by the spec, do you see any other purpose this library should produce keys for? Or was this parameter mostly added so that the negative case can be tested? 

Would it make sense to use a default parameter value?

aakoshh (commented):
Hats off to you, it's never easy to line up all the ducks when it comes to crypto APIs.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
KonradStaniec: So to be honest it was a mix of reading this rfc - https://tools.ietf.org/html/rfc5480, plus going through bouncy castle sources looking for the phrase `NamedCurve` or something similar 😅

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
KonradStaniec: So original spec does not say if this extension should be critical or not. In our project we always mark this extension as critical, but in theory someone could take `scalanet` library and modify it to connect to us without it being marked as criticial. So in theory it can happen, that we will get this extension as non critical one.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomTlsValidator.scala
KonradStaniec: I like it, it makes intent and the fact that we only use one extension more explicit

scalanet/src/io/iohk/scalanet/crypto/CryptoUtils.scala
KonradStaniec: Yep I mostly added it here to test negative case.

Original spec allows many possible combinations signatures and used hashes, it only requires that:
- hashes has at least 256bit
- RSA is avoided (but not forbidden)

I just restricted it here a little more as that way:
- it is easier to validate it
- I treat `scalanet` mostly as our project network layer not a generic network library as libp2p
- we already restrict some parts of the spec for our case, for example we only support secp256k1 keys in extension, where is original spec allows for 4 different kinds of keys.

My thinking goes, for now let's make everything as tailored to our use case as possible, and if we ever need to integrate with some software that uses libp2p tls connections, then we will extend our implementation to be as close to spec as possible.

And as far as default values, I try to avoid using them as they bit me several times in the past.

aakoshh (approved):
Looks good as far as I can tell! The docstrings really help understand what's going on.

PR #125, opened 2021-02-02
[FIX] Fix flaky tests

KonradStaniec:
# Description

Fixes some flaky tests in `KademliaIntegrationSpec`. Test failed due to the fact, in case of many nodes on single low-core machine, peers received `find-node` requests before they were properly bonded that lead to failing to accumulate enough peers to pass the test. Quick fix is to reduce the refresh rate, to do several refreshes before test end, then peers will ultimately be found.


aakoshh (commented):
> peers received find-node requests before they were properly bonded

So this happens with the v4 stuff, right? Thanks for fixing it.

aakoshh (approved):

KonradStaniec:
Yep this was happening with v4, there are some other flaky tests but those happened most often.

PR #126, opened 2021-02-04
[PM-2752] Throttle incoming connections from the same ip

KonradStaniec:
# Description

Create a possibility to throttle incoming connections from the same ip.

Some of design decisions:
- I did not add `maxSize` property to cache as it with typical throttling times (like lets say `30s`) it would be really hard to use up all host memory. (Ip4 address takes 4bytes, so 1M of them would take only 4Mb + some overhead for java object fields)
- I decided to use caffeine cache with `exipreAfterWrite` property. The similar effect could be accomplished by using java concurrent hash map, and scheduling clearing on netty eventLoops, but ultimately I think cache solution is clearer and worth bringing another dependency
- I decided to not give user of the group option to sent a message when closing the channel which violates throttling restrictions. By default we are treating connections which violates throttling restrictions as hostile and want to give it as little information as possilbe.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -150,7 +150,8 @@ class ScalanetModule(crossVersion: String) extends Module {
       ivy"org.scodec::scodec-core:1.11.7",
       ivy"org.bouncycastle:bcprov-jdk15on:1.64",
       ivy"org.bouncycastle:bcpkix-jdk15on:1.64",
-      ivy"org.bouncycastle:bctls-jdk15on:1.64"
+      ivy"org.bouncycastle:bctls-jdk15on:1.64",
+      ivy"com.github.ben-manes.caffeine:caffeine:2.8.8"
     )
 
     def scoverageVersion = "1.3.1"
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
@@ -8,25 +8,35 @@ import io.netty.handler.ipfilter.AbstractRemoteAddressFilter
 import java.net.{InetAddress, InetSocketAddress}
 
 private[scalanet] object CustomHandlers {
-  // to share handlers between pipelines they need to be marked as @Sharable, if not netty refuses to share it.
+
+  /**
+    *
+    * Custom handler which keeps recent history of incoming connections. If it receive new connection from the ip address
+    * which is still in history, it rejects it as it means the remote caller tries to often.
+    *
+    * Handler needs to be thread safe as it is shared between several netty pipelines
+    *
+    * To share handlers between pipelines they need to be marked as @Sharable, if not netty refuses to share it.
+    */
   @Sharable
   class ThrottlingIpFilter(config: DynamicTLSPeerGroup.IncomingConnectionThrottlingConfig)
       extends AbstractRemoteAddressFilter[InetSocketAddress] {
-    private val cache = Caffeine
+
+    private val cacheView = Caffeine
       .newBuilder()
       .expireAfterWrite(config.throttlingDuration.length, config.throttlingDuration.unit)
       .build[InetAddress, java.lang.Boolean]()
-    private val cacheView = cache.asMap()
+      .asMap()
 
-    private def addIfAbsent(address: InetAddress): Boolean = {
+    private def isQuotaAvailable(address: InetAddress): Boolean = {
       cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) == null
     }
 
     override def accept(ctx: ChannelHandlerContext, remoteAddress: InetSocketAddress): Boolean = {
       val address = remoteAddress.getAddress
-      val throttleLocalAddress = (address.isLoopbackAddress && !config.throttleLocalhost)
+      val localNoThrottle = (address.isLoopbackAddress && !config.throttleLocalhost)
 
-      throttleLocalAddress || addIfAbsent(address)
+      localNoThrottle || isQuotaAvailable(address)
     }
   }
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
@@ -1,0 +1,33 @@
+package io.iohk.scalanet.peergroup.dynamictls
+
+import com.github.benmanes.caffeine.cache.Caffeine
+import io.netty.channel.ChannelHandler.Sharable
+import io.netty.channel.ChannelHandlerContext
+import io.netty.handler.ipfilter.AbstractRemoteAddressFilter
+
+import java.net.{InetAddress, InetSocketAddress}
+
+private[scalanet] object CustomHandlers {
+  // to share handlers between pipelines they need to be marked as @Sharable, if not netty refuses to share it.
+  @Sharable
+  class ThrottlingIpFilter(config: DynamicTLSPeerGroup.IncomingConnectionThrottlingConfig)
+      extends AbstractRemoteAddressFilter[InetSocketAddress] {
+    private val cache = Caffeine
+      .newBuilder()
+      .expireAfterWrite(config.throttlingDuration.length, config.throttlingDuration.unit)
+      .build[InetAddress, java.lang.Boolean]()
+    private val cacheView = cache.asMap()
+
+    private def addIfAbsent(address: InetAddress): Boolean = {
+      cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) == null
+    }
+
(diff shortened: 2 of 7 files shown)


build.sc
```diff
@@ -150,7 +150,8 @@ class ScalanetModule(crossVersion: String) extends Module {
       ivy"org.scodec::scodec-core:1.11.7",
       ivy"org.bouncycastle:bcprov-jdk15on:1.64",
       ivy"org.bouncycastle:bcpkix-jdk15on:1.64",
-      ivy"org.bouncycastle:bctls-jdk15on:1.64"
+      ivy"org.bouncycastle:bctls-jdk15on:1.64",
+      ivy"com.github.ben-manes.caffeine:caffeine:2.8.8"
```
aakoshh, line 154: The projects we use Scalanet in already use `guava`, which also has a timed cache. Shouldn't we stick to that?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,33 @@
...
+private[scalanet] object CustomHandlers {
+  // to share handlers between pipelines they need to be marked as @Sharable, if not netty refuses to share it.
+  @Sharable
+  class ThrottlingIpFilter(config: DynamicTLSPeerGroup.IncomingConnectionThrottlingConfig)
+      extends AbstractRemoteAddressFilter[InetSocketAddress] {
+    private val cache = Caffeine
+      .newBuilder()
+      .expireAfterWrite(config.throttlingDuration.length, config.throttlingDuration.unit)
+      .build[InetAddress, java.lang.Boolean]()
+    private val cacheView = cache.asMap()
```
aakoshh:
```suggestion
    private val cacheView = Caffeine
      .newBuilder()
      .expireAfterWrite(config.throttlingDuration.length, config.throttlingDuration.unit)
      .build[InetAddress, java.lang.Boolean]()
      .asMap()
```
Would it make any difference? At least it would be obvious that `cache` isn't used for anything.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,33 @@
...
+
+    private def addIfAbsent(address: InetAddress): Boolean = {
+      cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) == null
+    }
```
aakoshh:
```suggestion
    private def isRateLimited(address: InetAddress): Boolean = {
      cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) != null
    }
```
The name `addIfAbsent` doesn't say much about the semantics of the cache. I thought something more suggestive would be better.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,33 @@
...
+
+    override def accept(ctx: ChannelHandlerContext, remoteAddress: InetSocketAddress): Boolean = {
+      val address = remoteAddress.getAddress
+      val throttleLocalAddress = (address.isLoopbackAddress && !config.throttleLocalhost)
+
+      throttleLocalAddress || addIfAbsent(address)
```
aakoshh:
```suggestion
      val localNoThrottle = (address.isLoopbackAddress && !config.throttleLocalhost)

      localNoThrottle || addIfAbsent(address)
```
The name `throttleLocalAddress` sounds exactly the opposite of `isLoopback && !throttleLocal`

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,33 @@
...
+import io.netty.channel.ChannelHandler.Sharable
+import io.netty.channel.ChannelHandlerContext
+import io.netty.handler.ipfilter.AbstractRemoteAddressFilter
+
+import java.net.{InetAddress, InetSocketAddress}
+
+private[scalanet] object CustomHandlers {
+  // to share handlers between pipelines they need to be marked as @Sharable, if not netty refuses to share it.
+  @Sharable
+  class ThrottlingIpFilter(config: DynamicTLSPeerGroup.IncomingConnectionThrottlingConfig)
```
aakoshh, line 22: I'd add some docstring explaining how throttling works in this class. The config explains it nicely, but when I initially saw at just this, and haven't yet looked up the the netty docs, I thought it might have been about limiting requests.

aakoshh (commented):
Nice one!

build.sc
KonradStaniec: hmm there is only some old versions of `guava` - `21`, which is from `12-Jan-2017`, and also it exports only some wierd annotation library part - `com.google.errorprone:error_prone_annotations`.  

Taking those into account I would stick with caffeine.

aakoshh: Maybe we should upgrade then, there's version 30 from Dec 2020.

Looking at [caffeine on Maven](https://mvnrepository.com/artifact/com.github.ben-manes.caffeine/caffeine/2.8.8) it has the same `error_prone_annotations` dependency, there's even an [issue](https://github.com/ben-manes/caffeine/issues/300) for it, which references the [guava issue](https://github.com/google/guava/issues/2824).

aakoshh: But it sounds like Caffeine is a bit better, like no extra thread for cleanups, so if you prefer it then :+1:

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,43 @@
...
+      extends AbstractRemoteAddressFilter[InetSocketAddress] {
+
+    private val cacheView = Caffeine
+      .newBuilder()
+      .expireAfterWrite(config.throttlingDuration.length, config.throttlingDuration.unit)
+      .build[InetAddress, java.lang.Boolean]()
+      .asMap()
+
+    private def isRateLimited(address: InetAddress): Boolean = {
+      cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) == null
```
aakoshh, line 32:
```suggestion
      cacheView.putIfAbsent(address, java.lang.Boolean.TRUE) != null
```
I thought if it wasn't null then it mean it was already in, so it has to be rate limited now.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    override def accept(ctx: ChannelHandlerContext, remoteAddress: InetSocketAddress): Boolean = {
+      val address = remoteAddress.getAddress
+      val localNoThrottle = (address.isLoopbackAddress && !config.throttleLocalhost)
+
+      localNoThrottle || isRateLimited(address)
```
aakoshh:
```suggestion
      localNoThrottle || !isRateLimited(address)
```
Ah, I see why the `==` didn't cause a problem :) 

I thought "rate limited" would mean that it hit the limit. Sorry maybe it's not a good name after all.

aakoshh: Maybe `isQuotaAvailable` is a better name, and then `true` means okay.

aakoshh (approved):
Looks good. I'm sorry for suggesting a confusing name, see my comment about another option.

build.sc
KonradStaniec: Sure we can,  I do not have strong preference here. As long it  has `putIfAbsent` and timed eviction it is good enough.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/CustomHandlers.scala
aakoshh: See my [other](https://github.com/input-output-hk/scalanet/pull/126#discussion_r570864482) comment about renaming the method to avoid double negation.

build.sc
aakoshh: The [CacheBuilder](https://guava.dev/releases/19.0/api/docs/com/google/common/cache/CacheBuilder.html) has an example of expire after write, and apparently the [asMap with putIfAbsent](https://stackoverflow.com/questions/48199370/how-to-figure-out-whether-a-key-exists-in-guava-cache-so-that-i-dont-overwrite) is still the way. 

I also read that Caffeine handles async loading natively, whereas Guava is synchronous.

KonradStaniec:
so ultimately lets go with `isQuotaAvailable` and caffeine cache :)

PR #127, opened 2021-02-08
[PM-1902] TLS 1.3

KonradStaniec:
# Description

Updates tls used to `TLS 1.3`. It seems tls 1.3 is now supported in both java8 and java11 (and newer versions) so when using jdk based tls there is no limits about jdk versions (except maybe to use newest possible version of java8). The details are in - https://github.com/netty/netty/commit/b1d3aad404a39143da7a86c121d60f35c0d21108

Also adds config option to use native versions of via statically linked jars with boring ssl. From various different netty presentations it seems that native ssl has 3x times larger throughput that jdk one.

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
@@ -261,7 +261,11 @@ object ReqResponseProtocol {
     val hostkeyPair = CryptoUtils.genEcKeyPair(rnd, Secp256k1.curveName)
 
     for {
-      config <- Resource.liftF(Task.fromTry(DynamicTLSPeerGroup.Config(address, Secp256k1, hostkeyPair, rnd, None)))
+      config <- Resource.liftF(
+        Task.fromTry(
+          DynamicTLSPeerGroup.Config(address, Secp256k1, hostkeyPair, rnd, useNativeTlsImplementation = false, None)
+        )
+      )
       pg <- DynamicTLSPeerGroup[MessageEnvelope[M]](config)
       prot <- buildProtocol(pg)
     } yield prot
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
@@ -5,7 +5,7 @@ import java.security.KeyStore
 import java.security.cert.X509Certificate
 import io.iohk.scalanet.peergroup.dynamictls.DynamicTLSPeerGroup.PeerInfo
 import io.netty.handler.ssl.util.SimpleTrustManagerFactory
-import io.netty.handler.ssl.{ClientAuth, SslContext, SslContextBuilder}
+import io.netty.handler.ssl.{ClientAuth, SslContext, SslContextBuilder, SslProvider}
 
 import javax.net.ssl._
 import scodec.bits.BitVector
@@ -73,11 +73,14 @@ private[scalanet] object DynamicTLSPeerGroupUtils {
   case class SSLContextForClient(to: PeerInfo) extends SSLContextFor
 
   def buildCustomSSlContext(f: SSLContextFor, config: DynamicTLSPeerGroup.Config): SslContext = {
+    val sslProvider = if (config.useNativeTlsImplementation) SslProvider.OPENSSL else SslProvider.JDK
+
     f match {
       case SSLContextForServer =>
         SslContextBuilder
           .forServer(config.connectionKeyPair.getPrivate, List(config.connectionCertificate): _*)
           .trustManager(new CustomTrustManagerFactory(None))
+          .sslProvider(sslProvider)
           .clientAuth(ClientAuth.REQUIRE)
           .protocols("TLSv1.3")
           .build()
@@ -87,6 +90,7 @@ private[scalanet] object DynamicTLSPeerGroupUtils {
           .forClient()
           .keyManager(config.connectionKeyPair.getPrivate, List(config.connectionCertificate): _*)
           .trustManager(new CustomTrustManagerFactory(Some(info.id)))
+          .sslProvider(sslProvider)
           .protocols("TLSv1.3")
           .build()
     }
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
@@ -3,22 +3,14 @@ package io.iohk.scalanet.peergroup.dynamictls
 import java.net.Socket
 import java.security.KeyStore
 import java.security.cert.X509Certificate
-
 import io.iohk.scalanet.peergroup.dynamictls.DynamicTLSPeerGroup.PeerInfo
 import io.netty.handler.ssl.util.SimpleTrustManagerFactory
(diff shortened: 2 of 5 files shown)


scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -261,7 +261,11 @@ object ReqResponseProtocol {
     val hostkeyPair = CryptoUtils.genEcKeyPair(rnd, Secp256k1.curveName)
 
     for {
-      config <- Resource.liftF(Task.fromTry(DynamicTLSPeerGroup.Config(address, Secp256k1, hostkeyPair, rnd, None)))
+      config <- Resource.liftF(
+        Task.fromTry(
+          DynamicTLSPeerGroup.Config(address, Secp256k1, hostkeyPair, rnd, useNativeTlsImplementation = false, None)
```
aakoshh, line 266: Why is it hardcoded to `false` for the `ReqResponseProtocol`?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
```diff
@@ -81,23 +73,25 @@ private[scalanet] object DynamicTLSPeerGroupUtils {
...
     f match {
       case SSLContextForServer =>
         SslContextBuilder
           .forServer(config.connectionKeyPair.getPrivate, List(config.connectionCertificate): _*)
           .trustManager(new CustomTrustManagerFactory(None))
+          .sslProvider(sslProvider)
           .clientAuth(ClientAuth.REQUIRE)
-          .ciphers(supportedCipherSuites)
-          .protocols("TLSv1.2")
+          .protocols("TLSv1.3")
```
aakoshh, line 85: After this change, an earlier client wouldn't be able to connect to a new one, right?

Is 1.2 still used on the ETC network?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -70,6 +70,42 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      case (server, ch1) =>
+        for {
+          _ <- ch1.sendMessage("Hello enc server")
+          rec <- server.serverEventObservable.collectChannelCreated.mergeMap {
+            case (ch, release) => ch.channelEventObservable.guarantee(release)
+          }.headL
+        } yield {
+          rec shouldEqual MessageReceived("Hello enc server")
+        }
+    }
```
aakoshh: These two tests look exactly the same, so you could have something like this:
```scala
for (nc <- List(true, false))
  for (ns <- List(true, false)) 
    it should s"handshake successfully using ${if (nc) "native" else "JDK"} client TLS and ${if (ns) "native" else "JDK"} server TLS" in {
       ...
    }
```

aakoshh (commented):
Looks good. 

Is this something we're doing in the Akka TCP connections as well? Is it okay to drop support for 1.2?

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
KonradStaniec: becouse `ReqResponseProtocol` is used only in tests, and for time being I do not expect it will be used anywhere else, thats why i am treating it a little bit neglectfully

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
KonradStaniec: yep, older clients using `DynamicTLSPeerGroup` with `TLS1.2` won't be able to connect to the ones using new version with `TLS1.3`. So it kinda breaking change.

ETC network does not use tls peer group, but have their own protocol called rlpx which is implemented in mantis ( I am guessing you are worried we would brake some mantis code ? ).

KonradStaniec:
This is something which we only doing here to be as most compatible as possible with libp2p specs (and they specify that tls1.3 should be forced).

The only reason we were not doing this until this time, is because when I had initially implemented this peergroup almost year ago there was no nice support of tls1.3 in netty, also tls1.3 was not implemented in java8 (so we would restrict ourselves to only newer java versions), so for time being I have decided to go with 1.2 and update it later when ecosystem will move forward a bit.

As for akka tcp connections (which are maintained betweend wallet and node) we could also force tls1.3 (as it is safer and faster than tls1.2) on those connections but it would be totally separate task.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
aakoshh: I knew it wasn't using Scalanet at the moment, I was asking more from a theoretical perspective: if they wanted to use it, could they? Or are we ruling out the use case by dropping support for 1.2?

KonradStaniec: I mean they could theoretically run tls1.3 connections between nodes along side with rlpx connections ( of course on other ports), but I do not see why would they do it.

aakoshh (commented):
Should we increment the version number? Since we're still in the 0.x series we can introduce breaking changes like this with just a minor version bump, but I noticed that you didn't do that in the previous PR either that added a new feature. I thought maybe you want to roll everything together and increment it once before release, so perhaps tag the PRs with the kind of change they imply?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
aakoshh: I literally meant: what if I wanted to use Scalanet to implement the network layer for a node that joins the Ethereum network?

KonradStaniec:
I planned to tag pr in our project repo, but sure i can tag it here also. As for versioning, my idea was to abandon constant version bumps, and just release library when necessary. As for now we are working on version `0.7.0` which in fact will contain several breaking changes. (at least in terms of tls connections, not discovery)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
KonradStaniec: then you would need to implement new peer group called something like `RlpxPeerGroup` that would be using encryption and framing used in ethereum. 
I remember proposing something like that when we were moving to mantis, but it was decided that it is not important for time being.

KonradStaniec: or you would need to make tls optional in this peergroup, and implement `Rlpx` on top of bare tcp connection

aakoshh: Hm, I thought our other project must have inherited the RLPx stuff too.

aakoshh (approved):
:lock:

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupUtils.scala
KonradStaniec: yep initially it inherited it, but it was deleted as soon as we have working tls connections between nodes

PR #128, opened 2021-02-12
[PM-2808] Change framing codec

KonradStaniec:
# Description

Changes home grown framing codec, to netty one which should be safer and more performant.

Codec are built for each channel from config provided by user of the peer group.


scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+        lengthFieldLength: Int,
+        encodingLengthAdjustment: Int,
+        decodingLengthAdjustment: Int,
+        initialBytesToStrip: Int,
+        failFast: Boolean = true,
+        byteOrder: ByteOrder = ByteOrder.BIG_ENDIAN,
+        lengthIncludesLengthFieldLength: Boolean = false
+    ): Option[FramingConfig] = {
+      val tooLargeOffset = lengthFieldOffset > maxFrameLength - lengthFieldLength
+      if (maxFrameLength <= 0 || lengthFieldOffset < 0 || initialBytesToStrip < 0 || !possibleLengthFieldLength
```
dmitry-worker: Perhaps `Option` here can be replaced with `Try` or `Either`, containing a self-explanatory error description.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+    private val possibleLengthFieldLength = Set(1, 2, 3, 4, 8)
+
+    /**
+      *
+      * Configures framing format for all the peers in PeerGroup
+      *
+      *
+      * @param maxFrameLength the maximum length of the frame. If the length of the frame is greater than this value
+      *                       channel with generate DecodingError event.
+      * @param lengthFieldOffset the offset of the length field.
```
dmitry-worker, line 206: Sorry, but it seems I cannot get the math of this offset. For example let's suppose a frame
`00 00 00 00 05` - can be decoded by framing config if 
1) `lengthFieldLength == 1`
2) `lengthFieldOffset == 4`
3) `lengthIncludesLengthFieldLength == true`
4) `maxFrameLength == 5`

And the data value in frame `00 00 00 00`. Is that correct?

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -256,14 +249,15 @@ object ReqResponseProtocol {
   def getTlsReqResponseProtocolClient[M](
       address: InetSocketAddress
   )(implicit s: Scheduler, c: Codec[M]): Resource[Task, ReqResponseProtocol[PeerInfo, M]] = {
-    implicit lazy val framingCodec = MessageEnvelope.framingCodec[M]
+    implicit lazy val envelopeCodec = MessageEnvelope.defaultCodec[M]
     val rnd = new SecureRandom()
     val hostkeyPair = CryptoUtils.genEcKeyPair(rnd, Secp256k1.curveName)
-
+    val framingConfig = FramingConfig.buildConfigWithStrippedLength(192000, 4).get
```
aakoshh: I know you're treating this class as dormant, but we still probably shouldn't add magic numbers like this. Maybe add a `FramingCodec.default` that can be used as a default argument in `getTlsReqResponseProtocolClient`, to at least make it configurable in case someone wants to use it for something larger than 192K? 

Why 192K?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
     }
   }
 
+  sealed abstract case class FramingConfig private (
```
aakoshh, line 148: I understand this prevents the creation of the default `apply` and `copy` constructors, but does it also prevent `unapply`? Would it be equivalent to `class FramingCodec private(val maxFrameLength: ..., val lenghtFieldOffset: ..., ...)` i.e. a normal class with all fields made public? In other words, what's the benefit of keeping it as a case class? Just curious.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+      *
+      * Configures framing format for all the peers in PeerGroup
+      *
+      *
+      * @param maxFrameLength the maximum length of the frame. If the length of the frame is greater than this value
+      *                       channel with generate DecodingError event.
+      * @param lengthFieldOffset the offset of the length field.
+      * @param lengthFieldLength  the length of the length field. Must be 1, 2, 3, 4 or 8 bytes.
+      * @param encodingLengthAdjustment the compensation value to add to the value of the length field when encoding.
+      * @param decodingLengthAdjustment the compensation value to add to the value of the length field when decoding.
```
aakoshh: I don't understand what these mean. Should they have default values?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+      * @param initialBytesToStrip the number of first bytes to strip out from the decoded frame. In standard framing setup
+      *                            it should be equal to lengthFieldLength.
+      * @param failFast  If true, a DecodingError event is generated as soon as the decoder notices the length of the frame will
+      *                  exceed maxFrameLength regardless of whether the entire frame has been read. If false,
+      *                  a a DecodingError event is generated after the entire frame that exceeds maxFrameLength has been read..
```
aakoshh:
```suggestion
      *                  a DecodingError event is generated after the entire frame that exceeds maxFrameLength has been read.
```
Why would you choose one or the other? What happens to the buffer if the error is raised fast?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+      * @param byteOrder the ByteOrder of the length field.
+      * @param lengthIncludesLengthFieldLength if true, the length of the prepended length field is added to the value of the prepended length field.
```
aakoshh, line 216: Wow, this "length" appears 3 times in this name!!!

It's interesting to think about it: if you have a length field with length 1, you can have content up to 255 bytes long, but if it's 255 bytes, and you add 1 for the prefix length, you get 256 which cannot be represented with the 1 byte prefix. Should that be an invalid configuration?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+  object FramingConfig {
+    private val possibleLengthFieldLength = Set(1, 2, 3, 4, 8)
+
+    /**
+      *
+      * Configures framing format for all the peers in PeerGroup
+      *
+      *
+      * @param maxFrameLength the maximum length of the frame. If the length of the frame is greater than this value
+      *                       channel with generate DecodingError event.
```
aakoshh, line 205: It should say whether it includes the prefix length or not, e.g. if my packet has a maximum length of 1280 bytes, and I want to use this encoding, should I add extra bytes for the prefix and set it to 1282?

Do we stream bytes, read the length-prefix and if it's over `maxFrameLength` it gets rejected immediately without the rest being read? Or does all the data _with_ the prefix have to fit in a frame of this length first, before parsing is attempted?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: I'd have thought the data comes after the length prefix. Maybe `initialBytesToStrip` should strip it? Otherwise there's little point in even including a field length, since you can only have 4 bytes before the length prefix, if that's your data then it has constant length.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: I agree wholeheartedly with `Either`.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+            lengthIncludesLengthFieldLength
+          ) {}
+        )
+      }
+    }
+
+    /**
+      *
+      * Configures framing format for all the peers in PeerGroup, which will pre-pend length to all messages when encoding
+      * and strip this length when decoding
```
aakoshh:
```suggestion
      * Configures framing format for all the peers in PeerGroup, which will prepend length to all messages when encoding
      * and strip this length when decoding.
```

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+        maxFrameLength: Int,
+        lengthFieldOffset: Int,
+        lengthFieldLength: Int,
+        encodingLengthAdjustment: Int,
+        decodingLengthAdjustment: Int,
+        initialBytesToStrip: Int,
+        failFast: Boolean = true,
+        byteOrder: ByteOrder = ByteOrder.BIG_ENDIAN,
+        lengthIncludesLengthFieldLength: Boolean = false
+    ): Option[FramingConfig] = {
```
aakoshh: I wonder if we really have to expose all these options, rather than just sensible defaults. I understand they are added because Netty supports them, but what if we decided not to use Netty for this after all, and these settings would not map clearly to whatever replaces it?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+      val tooLargeOffset = lengthFieldOffset > maxFrameLength - lengthFieldLength
```
aakoshh: So the `maxFrameLength` includes the prefix. Shouldn't there be a validation to check whether `maxFrameLength` can even be expressed with `lengthFieldLength` bytes?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,95 @@ object DynamicTLSPeerGroup {
...
+      * Configures framing format for all the peers in PeerGroup
+      *
+      *
+      * @param maxFrameLength the maximum length of the frame. If the length of the frame is greater than this value
+      *                       channel with generate DecodingError event.
+      * @param lengthFieldOffset the offset of the length field.
+      * @param lengthFieldLength  the length of the length field. Must be 1, 2, 3, 4 or 8 bytes.
+      * @param encodingLengthAdjustment the compensation value to add to the value of the length field when encoding.
+      * @param decodingLengthAdjustment the compensation value to add to the value of the length field when decoding.
+      * @param initialBytesToStrip the number of first bytes to strip out from the decoded frame. In standard framing setup
```
aakoshh, line 210: Would it be worth documenting explicitly what the standard is? Is it the other smart constructor? Maybe call that `buildStandardConfig` so people don't even have to think what `WithStrippedLength` means to them.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -228,9 +256,9 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
       serverQueue: CloseableQueue[ServerEvent[PeerInfo, M]],
```
aakoshh, line 256: Off topic: it would be so good to clean up the `ServerChannelBuilder` which is used as a side effecting constructor in `DynamicTLSPeerGroup.serverBootstrap`. Every time I look at this code I'm lost about what `ServerChannelImpl` and `ClientChannelImpl` are; only by tracking where they are created/called from can I put myself in the frame of figuring out which event is raised where and what the Netty channels represent.

Probably because in my mind the "server channel" is where incoming connections are accepted, whereas it just means "incoming client channel" vs "outgoing client channel". But at least if it was a static builder function it would indicate more clearly how its used. Maybe `ServerChannelImpl.build` or `apply`.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -314,6 +312,51 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
         }
     }
   }
+  it should "handle messages with prepended messages id-s" in taskTestCase {
```
aakoshh, line 312: I don't see from the test where these IDs are prepended, can you add some comments? Are you talking about request IDs, or a discriminator for message types (e.g. `Foo`)?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -314,6 +312,51 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+        } yield {
+          assert(eventReceived == MessageReceived(TestMessage.Foo("hello")))
+        }
+    }
+  }
+
+  it should "inform user about too large frame" in taskTestCase {
+    (for {
+      client <- DynamicTLSPeerGroup[Byte](getCorrectConfig())
+      // 1byte message + 4 bytes length field gives frame of size 5, so maxFrameLength = 4 should end with error
```
aakoshh: I assume these are the settings in `getCorrectConfig()`, right? 
Could you add them to the constructor here, so that if the defaults change this test doesn't get affected?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -314,6 +312,51 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      ch1 <- client.client(server.processAddress)
+    } yield (client, server, ch1)).use {
+      case (client, server, ch1) =>
+        for {
+          result <- Task.parZip2(
+            ch1.sendMessage(1.toByte),
+            server.serverEventObservable.collectChannelCreated.mergeMap {
+              case (channel, release) =>
+                channel.channelEventObservable.guarantee(release)
+            }.headL
```
aakoshh: Maybe this pattern could be added as an extension method somewhere in the test, it appears so often.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -324,10 +367,12 @@ object DynamicTLSPeerGroupSpec {
 
   def getCorrectConfig(
       address: InetSocketAddress = aRandomAddress(),
-      useNativeTlsImplementation: Boolean = false
+      useNativeTlsImplementation: Boolean = false,
+      maxFrameLength: Int = 192000
   ): DynamicTLSPeerGroup.Config = {
     val hostkeyPair = CryptoUtils.genEcKeyPair(rnd, Secp256k1.curveName)
-    Config(address, Secp256k1, hostkeyPair, rnd, useNativeTlsImplementation, None).get
+    val framingConfig = FramingConfig.buildConfigWithStrippedLength(maxFrameLength, 4).get
```
aakoshh: Move `4` to the constructor?

aakoshh (commented):
Cool, that was more straight forward then I thought, well done master!

The meaning of all the config values are not obvious though, I'm not convinced we need to expose the full flexibility of Netty right away and push it all on the user.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: personally, I do not see much gain from `Either[String, FramingCodec]` vs `Option[FramingCodec]`, especially as most of most of those checks just checks are pretty self explanatory and if something goes wrong it is pretty easy to check implementation. 

but I am overvoted here so lets go this way :+1:

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: Yep standard case is `length + message` , and then in decoder we are stripping length bytes.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
KonradStaniec: We can make it as default argument to `getTlsReqResponseProtocolClient` but i would avoid adding something like `FramingCodec.default` as `maxFrameLength` is something really per protocol/peergroup basis, so making default case makes little sense imo. Nevertheless I will get rid of this hardcoded `4`

KonradStaniec: and 192k becouse it is nice even number. This is only used in tests so whatever would be good here

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: I think benefits here are nice default to `toString` and `equals` plus `hashCode`. I am not sure why you would prevent `unapplay` to prevent pattern matching ? (from what can be seen here `https://gist.github.com/tpolecat/a5cb0dc9adeacc93f846835ed21c92d2, unapplay still works`)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: My take on that is that at this point, it is more probable that we will fiddle with framing format than change natty itself, therefore I thought to have one constructor that expose everything (in case it will be needed) and second constructor (`buildConfigWithStrippedLength`) with sensible defaults with framing format which we use right now.

So in case in future we will do something not standard in framing department we will be able to do it without bumping library version.

scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
aakoshh: Okay, fair point about `FramingCodec.default`, I'm happy with hardcoding to the `ReqResponseProtocol` as long as it can be overridden.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: Ah, cool, I forgot about equality, that makes sense. I didn't want to prevent pattern matching, I was just curious if it worked (happy to see that it does). Thanks for the link!

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: from what I know those checks are done when encoding and decoding messages. The only checks I have added here are the checks which netty does inside their constructors i.e it will always be possible to construct netty encoder and decoder from properly constructed `FramingConfig`

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: The only reason I can imagine you wanting to change it is if the other side had some obscure format that doesn't match the defaults. I think in at least some codecs we had: `length-of-length-of-content || length-of-content || content`, so the actual length prefix was dynamic. I'm not sure if these config values could capture that, since while you know the offset of the length-prefix, it doesn't have a fixed length. 

Fine though, it's no problem to ignore these. I agree we're not getting rid of Netty; it's more likely people would rather want to switch away from Scalanet to https://github.com/typelevel/fs2-netty than change Netty.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: I mean if I created a config with `maxFrameLength = 65535, lengthFieldLength = 1, lengthFieldOffset = 0`, then you could not capture the maximum length with the available bits in the length field. I don't know if netty constructors would complain, but this condition doesn't capture it.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -27,11 +26,35 @@ import scodec.bits.BitVector
 import scala.concurrent.Promise
 import scala.util.control.NonFatal
 import io.netty.channel.EventLoop
+import io.netty.handler.codec.{LengthFieldBasedFrameDecoder, LengthFieldPrepender, TooLongFrameException}
 import io.netty.handler.proxy.Socks5ProxyHandler
+import scodec.Attempt.{Failure, Successful}
+import scodec.Codec
 
 private[dynamictls] object DynamicTLSPeerGroupInternals {
+  def buildFramingCodecs(config: FramingConfig): (LengthFieldBasedFrameDecoder, LengthFieldPrepender) = {
```
aakoshh, line 35: If you say the Netty constructors do checks and raise exceptions, maybe they should be constructed in a `Try`?

Better yet, have some property based tests to see that no combination of parameters can result in a config that passes your checks and then fails Netty.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: From what i know, in both cases buffer will discarded. In normal case it will be done after reading whole buffer, in `failFast` case it will be done as soon as violation is discvoered and after throwing `TooLongFrameException` decoder will enter byte discarding mode i.e it will discard next upcoming necessary bytes.

About choosing one or another, I am not sure. Main difference is that in fail fast mode downstream handlers will not get information about how big this violating frame would be. So there may be some protocols to react those violations.  ( we are not doing any of that thats why default is `failFast= true`)

aakoshh: Yes we don't say anything in `DecodingError`. Right so Netty would drop as many bytes as the stream indicated, that makes sense, thanks!

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: from netty perspective it will fail when encoding. I kinda agrees with this approach, as there may be the messages which are smaller than limit which will works just fine

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: as I said  in some comment, if constructed from properly created `FramingConfig` they will be no exceptions i.e `FramingConfig` does the same checks as netty codecs.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: Seems like good idea, although I would to this after my next pr which touches a little bit this area. (I will create task there or some kinda TODO)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: so situation here is that constructors would not complain. The failure would happen when trying to encode message which would violate this `lengthFieldLength` field i.e all would be fine when smaller messages would be sent, and only sending messages larger than 255 would fail

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: oh nice i did not know such projects existed, it is definitly worth keeping an eye on it

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: So first question depends on `lengthIncludesLengthFieldLength` flag  i.e. if it is `true` then it is true that prepended length is part of the `maxFrameLength`.

And second question depends on `failFast` flag. if its `true` then it is happening as you described if `false`  then the whole thing is read and then checked

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: Yep, I will add a little explanation plus default values.

KonradStaniec: Ultimatly, I have linked to netty docs as they are much more comprehensive

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: Hm, then I may have misunderstood `lengthIncludesLengthFieldLength`. 

What I thought it meant was that for example:
1. I configure a `lengthFieldLength` of 4
2. I'm reading the stream of data
3. I read 4 bytes
4. I evaluate those bytes to `length`
5. Because `lengthIncludesLengthFieldLength` is true, I assume it will be 4 higher than what the length of the data that follows.
6. I read `length - 4` bytes

Whereas if `lengthIncludesLenghtFieldLength` was false, I'd assume that it goes:
...
5. Because `lengthIncludesLengthFieldLength` is false, I assume it will be only contain the length of the actual content
6. I read `length` bytes

I thought that in both cases, the `maxFrameLength` I have to configure has to cater for the data as well as the prefix.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: Then I suggest adding these to the validations you do :)

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Thanks :+1:

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
aakoshh: Okay fair enough, then there's no point in property based testing. The fact that encodings can raise exceptions even with configuration that this accepts is more worrying.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: oh maybe I have said somehing not precise enough. So in case of option `lengthIncludesLengthFieldLength` it works only on `encoder` i.e it is used only in `LengthFieldPrepender` class, and when `true` length prefix is increased by length prefix length. 

On decoder side i.e in `LengthFieldBasedFrameDecoder`, in that case one need to configure option `decodingLengthAdjustment` to be inverse of length field length to properly account from it with `maxFrameLength`
from netty docs:
```
In most cases, the length field represents the length of the message body only, as shown in the previous examples. However, in some protocols, the length field represents the length of the whole message, including the message header. In such a case, we specify a non-zero lengthAdjustment. Because the length value in this example message is always greater than the body length by 2, we specify -2 as lengthAdjustment for compensation.
   lengthFieldOffset   =  0
   lengthFieldLength   =  2
   lengthAdjustment    = -2 (= the length of the Length field)
   initialBytesToStrip =  0
```

aakoshh: I tap out :upside_down_face: 

I put all my hope that the standard use case is simple and that's all anyone ever has to use, or else take a deep dive in Netty docs.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -141,6 +142,102 @@ object DynamicTLSPeerGroup {
...
+        initialBytesToStrip: Int,
+        encodingLengthAdjustment: Int = 0,
+        decodingLengthAdjustment: Int = 0,
+        failFast: Boolean = true,
+        byteOrder: ByteOrder = ByteOrder.BIG_ENDIAN,
+        lengthIncludesLengthFieldLength: Boolean = false
+    ): Either[ConfigError, FramingConfig] = {
+      val smallEnoughOffset = lengthFieldOffset <= maxFrameLength - lengthFieldLength
+      for {
+        _ <- Either.cond(maxFrameLength > 0, (), ConfigError("maxFrameLength should be positive"))
```
aakoshh: A small helper method could turn these into something like `check(maxFrameLength > 0, "maxFrameLength should be positive")`

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
aakoshh: Sorry, I meant adding them to the `client <- DynamicTLSPeerGroup[Byte](getCorrectConfig(maxFrameLength=5, lenghtFieldLength=4))` so that the what the comment says are not just assumption knowing what the test setup is, but always forces it to be what this test needs.

aakoshh (approved):
Awesome, thanks for the extra validation and calculations, that should help!

KonradStaniec:
sure np,  thanks for insisting on it ultimately pr looks much better after round of comments :)

PR #129, opened 2021-02-15
PM-2720: Discern peers by network

aakoshh:
# Description

Adds the ability to filter the peers before storing them for routing by the contents of their ENR records. Currently the filtering is in the form of tags, which are symmetric: if we set a tag on the outgoing ENR, and we expect it to be present in the incoming ENR. The mechanism allows setting attributes without imposing any filtering logic, or filtering without adding any attributes.

The caveat is that this is not checked during recursive lookups, only in the ENR fetches going on in the background. So if we're using the periodic background lookups, followed by the occasional `getNodes`, then everything is fine, all nodes returned will be in the same network as us. However if we use the new `getClosestNodes`, then it may first get a shortlist of closest nodes, and _then_ realise that they aren't on our network, and return less than 16 peers. I didn't want to change things too much, adding further latency to the recursive lookups (would need to check every neighbour returned), when this could only be an inconvenience if someone's maliciously trying to pollute the results with nodes from a different network. These would have to be closer to the target than anything on this network and respond to pings as well, it's not a very effective attack.

Another option would have been to try to add network IDs to the packets, but that requires lots of changes since the codecs are currently defined in Mantis (the RLP library hasn't been published yet).

# Important Changes Introduced

Mantis is doing random lookups to speed up the discovery process, however when I exposed `lookup` and `lookupRandom` I forgot that these don't go through the ENR fetch, the results are returned from the `Neighbors` responses. This is how the spec works, so this isn't a bug, but there's a slight vulnerability there: the TCP port is not checked, so an attacker could send garbage there. This would be inconvenient since Mantis could not establish the connection, but once the ENR record is retrieved we'd get the correct value.

Nevertheless I made `lookup` and `lookupRandom` private again, and instead added new `getRandomNodes` and `getClosestNodes` methods, which do make sure that we're only returning results for which we have an ENR as well.


scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
```diff
@@ -309,6 +325,20 @@ object DiscoveryService {
         }
       }
 
+    /** Perform a lookup and also make sure the closest results have their ENR records fetched,
+      * to rule out the chance that incorrect details were relayed in the Neighbors response.
+      */
+    override def getClosestNodes(target: Node.Id): Task[SortedSet[Node]] =
+      for {
+        closest <- lookup(target)
+        _ <- closest.toList.traverse(n => maybeFetchEnr(toPeer(n), None))
```
KonradStaniec: couldn't it be `parTraverse` to avoid waiting for results after eeach query ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/KeyValueTag.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    private val valueBytes =
+      ByteVector(value.getBytes(UTF_8))
+
+    override val toAttr =
+      Some(keyBytes -> valueBytes)
+
+    override val toFilter = enr =>
+      enr.content.attrs.get(keyBytes) match {
+        case Some(`valueBytes`) =>
```
KonradStaniec: just personal opinion, I really dislike `` matching, imo it would be more readable with standard `case Some(value) if value == valueBytes`

KonradStaniec (commented):
Only small comments.

Just a general question (as I not sure about some implementation details), with this apprach we will still respond to `ping`, `findNodes` from other networks, also we will able to bond with the. The one thing that changes, is that we will not store those peers from other networks in our routing table ? Do I understand it correctly ?

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Good idea! I was only in the mindset that these should be available already, or being fetched, but you're right if they aren't that's a new fetch attempt and then a timeout applies.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/KeyValueTag.scala
aakoshh: Hm, okay, I though tit was kind of cool not to have to name the variable if we don't need them, and it could scale to more occurrences too. Can change it back.

aakoshh:
Yes, correct, the protocol messages don't change, we still allow pinging and bonding initially, but we'd not add peers from other networks to the K-table, not even the bootstrap nodes, so we never spread this kind of info.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/KeyValueTag.scala
aakoshh: Done.

scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala
aakoshh: Done.

KonradStaniec (approved):

PR #130, opened 2021-02-17
[PM-2807] Tls back pressure

KonradStaniec:
# Description

Add possibility of stopping reading from RCV buffer when internal incoming queue is empty. Few caveats and design constraints:

- `incomingMessageQueue` may sometimes go out of bounds if netty will read data of more than one message from RCV buffer
-  Currently we are using default adaptive strategies form RCV and SND buffers sizes. We may consider adding options to size them by the user. Interesting options are : ` SO_SNDBUF`, `SO_RCVBUF` and `WRITE_BUFFER_WATER_MARK`. They decide when change channel writeablitiy status. We can do follow up task to expose them to the user.
-  Currently when all buffers are full,  `channel.sendMessage` returns `Task` which won't finish until receiver will consume some data. It is up to to user of channel (and peer group)  to decide if he want't to add some `timeout` on this `Task` and what to do in case of timeout. 
-  In our downstream project, we queue messages to send until previous one do not end in SND buffer, so we can easily adapt to this changes i.e have max outgoing queue size, and react accordingly to certain flushes progressing to slow. Ultimatly we could add outgoing queue on scalanet size and inform user when it is full (netty event loop thread would check this queue and writeAndFlush messages as they arrive)
- We could add check `channel.isWriteable` before running `channel.writeAndFlush` in `sendMessage`, and return failed task to user when it is false. This would require adding a bit coordination between threads to avoid race conditions, so I am still not sure if its good trade-off


scalanet/src/io/iohk/scalanet/peergroup/ReqResponseProtocol.scala
```diff
@@ -230,7 +230,7 @@ object ReqResponseProtocol {
       config <- Resource.liftF(
         Task.fromTry(
           DynamicTLSPeerGroup
-            .Config(address, Secp256k1, hostkeyPair, rnd, useNativeTlsImplementation = false, framingConfig, None)
+            .Config(address, Secp256k1, hostkeyPair, rnd, useNativeTlsImplementation = false, framingConfig, 100, None)
```
aakoshh: A named argument would be great here, I have no idea what 100 is for.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+/**
+  * Wraps an underlying unbounded CloseableQueue queue and bounds it based on netty auto-read feature.
+  * While auto-read is disabled received messages start to accumulate in underlying os RCV tcp buffer. When RCV buffer is full
+  * sender SND buffer will start to buffer un-sent bytes. When sender SND buffer is full, the default behaviour is that
+  * write(xxx) will block.  In our case, sendMessage Task will not finish until there will be place in SND buffer
+  *
+  *
+  * WARNING: Actual limit may sometimes goes higher, as each netty read can return more than one message.
```
aakoshh:
```suggestion
  * WARNING: Actual limit may sometimes go higher, as each netty read can return more than one message.
```

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
```diff
@@ -0,0 +1,63 @@
...
+  * @param queue is the underlying closeable message queue
+  */
+private[scalanet] final class ChannelAwareQueue[M] private (limit: Int, queue: CloseableQueue[M]) {
+  private val queueSize = new AtomicLong(0)
+
+  private val lowerBound: Int = Math.max(1, limit / 2)
+
+  def size: Long = queueSize.get()
+
+  def offer(channelConfig: ChannelConfig, a: M): Task[Unit] = {
```
aakoshh: Why isn't the return type of `CloseableQueue.offer` preserved?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
```diff
@@ -0,0 +1,63 @@
...
+    Task(enableBPIfNecessary(channelConfig)).flatMap(_ => queue.offer(a).void)
```
aakoshh:
```suggestion
    Task(enableBPIfNecessary(channelConfig)) >> queue.offer(a)
```

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def next(channelConfig: ChannelConfig): Task[Option[M]] = {
```
aakoshh: Couldn't you move `channelConfig` into the constructor in the companion? Does it change during the liftetime? That would be bad, because it would mean that `offer` can disable one channel and `next` enable a different one, which doesn't make sense, does it?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -113,30 +119,26 @@ private[dynamictls] object DynamicTLSPeerGroupInternals {
...
-      extends Channel[PeerInfo, M]
-      with StrictLogging {
+      extends StrictLogging {
     val (decoder, encoder) = buildFramingCodecs(framingConfig)
-
-    val to: PeerInfo = peerInfo
-
-    private val activation = Promise[io.netty.channel.Channel]()
+    private val to = peerInfo
+    private val messageQueue = makeMessageQueue[M](maxIncomingQueueSize)
```
aakoshh: I see that `ch.channelConfig()` first becomes available in `initChannel`, but you need the `messageQueue` in `initialize`. So maybe, like the `activation`, this needs to be put in a `Deferred` or `Promise`, and created in `initChannel` where you can pass the `channelConfig` to its constructor. Then `initialize` can take it from there by awaiting the promise.

aakoshh (commented):
Good job! I haven't looked at all the tests yet, but I have some comments regarding the way config is passed to the queue.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
KonradStaniec: because we never make us of it i.e in all call sites we were just voiding it. If we will ever need to expose full return type from `ChannelAwareQueue `for some reason it will be easy to bring it back.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
KonradStaniec: tbh In general i am not great fan of additional symbols (I always forget what they mean and need to go over the docs) but lets say this one is ubiquitous enough
My red flag is when I need to import whole `cats.implicits._` scope to to use some symbol.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
KonradStaniec: You are right, I just wanted avoid code reorganisation which you described in next comment. Potential for misuse is pretty small here as we are using this queue in one context only, so I thought we can live without it.  

But sure i can go that way 👍

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
aakoshh: Yes but there's no reason for every queue abstraction to have an ever so slightly different interface. When it even says in the docs that this is a wrapper over that other type, they should be interchangeable.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
aakoshh: You can be more specific by importing from `cats.syntax` or `cats.instances` but that requires more familiarity, see [here](https://typelevel.org/cats/typeclasses/imports.html) and [here](https://typelevel.org/cats/jump_start_guide.html).

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
aakoshh: Yeah I think it's better to have the right abstraction and massage the code than admit that it's so wrecked that we just need to pile on.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
KonradStaniec: you can, but imo it is bad trade off with most of the symbols, as you force this familiarity onto future maintainers (and upon yourself) in most cases just to save few keystrokes.

aakoshh: That's why I don't do it, and also because of the implicit resolution conflicts. `cats.implicits._` gives you everything (like `.void` for that matter).

It's not the key strokes I think are worth saving here, it just looks and reads much nicer. I don't know esoteric ones like `><>` whatever but some basics like `*>` and `<*` are great to clean up the code a bit.

scalanet/src/io/iohk/scalanet/peergroup/InetPeerGroupUtils.scala
```diff
@@ -1,29 +1,47 @@
...
 import java.net.{InetSocketAddress, ServerSocket}
-
-import io.netty.util
+import io.netty
+import io.netty.util.concurrent.{Future, GenericFutureListener}
 import monix.eval.Task
 
+import java.util.concurrent.CancellationException
+
 object InetPeerGroupUtils {
```
aakoshh: This class is getting ever more unrelated utility methods, some of which should be moved to tests (`aRandomNode`) and now most of which are about netty futures. Would be good to clean it up a bit, make it private to scalanet.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -349,6 +375,142 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
         }
     }
   }
+
+  it should "not send another message when receive queue and both buffers are full " in backPressureTestCase(
+    clientConfig = getCorrectConfig(maxFrameLength = 64 * 1024 * 1024),
+    serverConfig = getCorrectConfig(maxQueueSize = 1, maxFrameLength = 64 * 1024 * 1024)
+  ) {
+    case (clientChannel, serverChannel) =>
+      // we are using pretty large messages to be sure incoming and sending buffers will be full
```
aakoshh, line 384: What's the sending buffer size limit?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
```diff
@@ -0,0 +1,68 @@
...
+        disableBPIfNecessary()
+        Some(value)
+      case None =>
+        None
+    }
+  }
+
+  def close(discard: Boolean): Task[Unit] = queue.close(discard)
+
+  private def enableBPIfNecessary(): Unit =
```
aakoshh: What's "BP" stand for?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -349,6 +375,142 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+  }
+
+  it should "get more bytes form rcv buffer as soon as incoming queue will be at least half empty" in backPressureTestCase(
+    clientConfig = getCorrectConfig(maxFrameLength = 64 * 1024 * 1024),
+    serverConfig = getCorrectConfig(maxQueueSize = 4, maxFrameLength = 64 * 1024 * 1024)
+  ) {
+    case (clientChannel, serverChannel) =>
+      val srvChannelImpl =
+        serverChannel.asInstanceOf[DynamicTLSPeerGroupInternals.DynamicTlsChannel[ChannelEvent[SizeAbleMessage]]]
+      // we are using small messages to not fill up snd and rcv buffers
```
aakoshh, line 457: Can these be configured explicitly?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -349,6 +375,142 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+        // last two messages will be stored in rcv buffer, becouse incoming queue is full
+        _ <- Task.traverse(secondBatch)(m => clientChannel.sendMessage(m))
+        _ <- Task.sleep(1.seconds)
+        _ <- Task(assert(srvChannelImpl.incomingQueueSize == firstBatch.size))
+        _ <- srvChannelImpl.nextChannelEvent.map(ev => assert(ev.get == MessageReceived(firstBatch(0))))
+        _ <- Task(assert(srvChannelImpl.incomingQueueSize == firstBatch.size - 1))
+        _ <- srvChannelImpl.nextChannelEvent.map(ev => assert(ev.get == MessageReceived(firstBatch(1))))
+        // after reading first 2 messages, there place in queue for two new messages, so they are read from rcv buffer
+        _ <- Task(srvChannelImpl.incomingQueueSize)
+          .restartUntil(qSize => qSize == ((firstBatch.size - 2) + secondBatch.size))
```
aakoshh, line 474: What does this achieve?

aakoshh: Ah, it's like putting `eventually` around the next assertion.

aakoshh: If this condition never happened, would the whole test just time out? Maybe it would be worth putting some timeout on this explicitly, so we get a bit more clue about where the test stopped.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -349,6 +375,142 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+
+  it should "communicate with slower consumer with small queue available" in backPressureTestCase(
+    clientConfig = getCorrectConfig(maxFrameLength = 64 * 1024 * 1024),
+    serverConfig = getCorrectConfig(maxQueueSize = 1, maxFrameLength = 64 * 1024 * 1024)
+  ) {
+    case (clientChannel, serverChannel) =>
+      val srvChannelImpl =
+        serverChannel.asInstanceOf[DynamicTLSPeerGroupInternals.DynamicTlsChannel[ChannelEvent[SizeAbleMessage]]]
+      // large message so fill the buffer after each sent
+      val messagesToSend = (0 to 24).map(_ => SizeAbleMessage.genRandomOfsizeN(4 * 1024 * 1024))
```
aakoshh:
```suggestion
      val messagesToSend = List.fill(24)(SizeAbleMessage.genRandomOfsizeN(4 * 1024 * 1024))
```

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
```diff
@@ -349,6 +375,142 @@ class DynamicTLSPeerGroupSpec extends AsyncFlatSpec with BeforeAndAfterAll {
...
+        .toListL
+      for {
+        receivedMessages <- Task.parMap2(sendTask, receiveTask)((_, received) => received)
+        _ <- Task(assert(messagesToSend == receivedMessages))
+        finalQueueSize <- Task(
+          serverChannel.asInstanceOf[DynamicTLSPeerGroupInternals.DynamicTlsChannel[_]].incomingQueueSize
+        )
+      } yield {
+        assert(finalQueueSize == 0)
+      }
```
aakoshh, line 512: What is this testing exactly? It looks like it's checking that the receiver side has consumed all messages from the queue. Since you explicitly `.take(messagesToSend.size)` items from it, what else would the final count be other than 0?

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
KonradStaniec: It is really os/tcp implementation dependant, but in case of tests which fills buffers we are using messages large enough to fill them in all cases.

Those config could be exposed externally (as I described in pr description) by:

> Currently we are using default adaptive strategies form RCV and SND buffers sizes. We may consider adding options to size them by the user. Interesting options are : SO_SNDBUF, SO_RCVBUF and WRITE_BUFFER_WATER_MARK. They decide when change channel writeablitiy status. We can do follow up task to expose them to the user.

But even then options `SO_SNDBUF` and `SO_RCVBUF` are only hints to OS, and in some cases OS doubles this values, so we can't relay on them exactly in tests

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
KonradStaniec: It stands for `BackPressure`, I will add full naming here

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
KonradStaniec: It is as I have explained in other comment

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
KonradStaniec: good idea ! :+1:

aakoshh (approved):
This looks correct, great stuff! 

My comments are mostly asking for clarifications.

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
KonradStaniec: It is checking that explicit queue size counter in `ChannelAwareQueue` is decremented and incremented correctly.  If we would forgot to decrement it after consuming message then this test would fail

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/ChannelAwareQueue.scala
aakoshh: Ah, yes, it's in the title :man_facepalming:

scalanet/ut/src/io/iohk/scalanet/peergroup/DynamicTLSPeerGroupSpec.scala
aakoshh: Right, thanks, I forgot about the PR description.

PR #131, opened 2021-02-22
Release/0.7.0

KonradStaniec:
# Description

Release version 0.7.0

# Features added since 0.6.0

#124 - add new validations from libp2p specs
#126 - add option to throttle incoming connections from the same ip
#127 - force tls1.3 on tls connections
#128 - use netty native framing codecs instead of homegrown ones
#129 - discerns peers in different networks on discovery level
#130 - add basic back pressoure for incoming messages by using netty auto read feature


aakoshh (approved):
:+1:

aakoshh:
Maybe we should have a CHANGELOG, or just list what was added in these PRs.

KonradStaniec:
good point! for now I have added list of changes and pr-s merged since last release. I will add change log in next release iteration

aakoshh:
That list is perfect, thanks!

PR #134, opened 2021-04-12
[PM-3177] Add api to get channel local info

KonradStaniec:
# Description

Exposes local channel info for network channel, which may be usefull in some cases, for example, checking ephemeral port attached to outgoing `DynamicTlsPeerGroup` channel


aakoshh (approved):
Nice one!

PR #135, opened 2021-04-16
[Pm-3178] Add idle peer detection to DynamicTlsPeerGrup

KonradStaniec:
# Description

Adds idle peer detection to DynamicTlsPeerGrup by using netty `IdleStateHandler`. This makes it possible to implement different stale peer detection schemes on top of this peer group.

Simplest one for static cluster would be to require all nodes in cluster to sends heart beats every `x` seconds, and configure `readerIdleTime` to the same `x` seconds, then number of   `ChanneIdle(ReaderIdle)` events will represent number of missed heartbeats from remote node. (of course this is only approximation, in reality, due to set of different factors heartbeats could be delayed or queued somewhere)

### DIFF
--- a/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
@@ -76,7 +76,7 @@ object Channel {
     *  2. `first` flag means that this is the first idle event of given type since last activity i.e
     *       - in case `ReaderIdle -> AllIdle -> ReaderIdle`, the idle events look like:
     *       `ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(AllIdle, first = true) -> ChannelIdle(ReaderIdle, first = false)`
-    *       - in case `ReaderIdle -> IncomingRead -> ReaderIdle -> ReaderIdle`
+    *       - in case `ReaderIdle -> IncomingRead  -> ReaderIdle -> ReaderIdle`
     *       `ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(ReaderIdle, first = false)`
     *
     * @param idleState type of idle event
--- a/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
@@ -67,7 +67,22 @@ object Channel {
   final case class UnexpectedError(e: Throwable) extends ChannelEvent[Nothing]
   final case object DecodingError extends ChannelEvent[Nothing]
 
-  // only used in DynamicTlsPeerGroup
+  /**
+    *  only used in DynamicTlsPeerGroup, indicated that there was no action on particular channel
+    *  Basic rules regarding idle events:
+    *  1. Idle events are raised once per timeout i.e if reader idle time is configured to 5 second, then every
+    *     5 seconds ChannelIdle(ReaderIdle, _) event will be generated if there is no incoming traffic
+    *
+    *  2. `first` flag means that this is the first idle event of given type since last activity i.e
+    *       - in case `ReaderIdle -> AllIdle -> ReaderIdle`, the idle events look like:
+    *       `ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(AllIdle, first = true) -> ChannelIdle(ReaderIdle, first = false)`
+    *       - in case `ReaderIdle -> IncomingRead -> ReaderIdle -> ReaderIdle`
+    *       `ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(ReaderIdle, first = true) -> ChannelIdle(ReaderIdle, first = false)`
+    *
+    * @param idleState type of idle event
+    * @param first flag indicating if this is first idle event since last activity
+    *
+    */
   final case class ChannelIdle(idleState: IdleState, first: Boolean) extends ChannelEvent[Nothing]
 }
 
--- a/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
@@ -57,10 +57,18 @@ trait Channel[A, M] {
 }
 
 object Channel {
+  sealed abstract class IdleState
+  case object ReaderIdle extends IdleState
+  case object WriterIdle extends IdleState
+  case object AllIdle extends IdleState
+
   sealed abstract class ChannelEvent[+M]
   final case class MessageReceived[M](m: M) extends ChannelEvent[M]
   final case class UnexpectedError(e: Throwable) extends ChannelEvent[Nothing]
   final case object DecodingError extends ChannelEvent[Nothing]
+
+  // only used in DynamicTlsPeerGroup
+  final case class ChannelIdle(idleState: IdleState, first: Boolean) extends ChannelEvent[Nothing]
 }
 
 /**
--- a/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
+++ b/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
@@ -33,7 +33,7 @@ import scodec.Codec
 import scodec.bits.BitVector
 
 import java.nio.ByteOrder
-import scala.concurrent.duration.{FiniteDuration, TimeUnit}
+import scala.concurrent.duration.FiniteDuration
 import scala.util.Try
 import scala.util.control.NonFatal
 
@@ -312,7 +312,6 @@ object DynamicTLSPeerGroup {
 
   /**
     * Configures detection on idle peer
-    * @param timeUnit unit of time for all time configs
     * @param readerIdleTime a ChannelIdle event whose state is ReaderIdle will be triggered when
     *                       no read was performed for the specified period of time on given channel. Specify 0 to disable.
     * @param writerIdleTime a ChannelIdle event whose state is WriterIdle will be triggered when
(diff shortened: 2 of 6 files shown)


scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
```diff
@@ -57,10 +57,18 @@ trait Channel[A, M] {
 }
 
 object Channel {
+  sealed abstract class IdleState
+  case object ReaderIdle extends IdleState
+  case object WriterIdle extends IdleState
+  case object AllIdle extends IdleState
```
aakoshh, line 63: Just thinking: would it be easier to represent these with two flags?
```scala
case class ChannelIdle(readerIdle: Boolean, writerIdle: Boolean, first: Boolean) extends ChannelEvent[Nothing]
```

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -56,10 +59,6 @@ class DynamicTLSPeerGroup[M] private (val config: Config)(
 
   private val sslServerCtx: SslContext = DynamicTLSPeerGroupUtils.buildCustomSSlContext(SSLContextForServer, config)
 
-  // Using MPMC because the channel creation event is only pushed after the SSL handshake,
-  // which should take place on the channel thread, not the boss thread.
-  private val serverQueue = CloseableQueue.unbounded[ServerEvent[PeerInfo, M]](ChannelType.MPMC).runSyncUnsafe()
```
aakoshh, line 61: :love_you_gesture:

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -309,6 +310,53 @@ object DynamicTLSPeerGroup {
...
 
+  /**
+    * Configures detection on idle peer
+    * @param timeUnit unit of time for all time configs
+    * @param readerIdleTime a ChannelIdle event whose state is ReaderIdle will be triggered when
+    *                       no read was performed for the specified period of time on given channel. Specify 0 to disable.
+    * @param writerIdleTime a ChannelIdle event whose state is WriterIdle will be triggered when
+    *                       no write was performed for the specified period of time on given channel. Specify 0 to disable.
+    * @param allIdleTime a ChannelIdle event whose state is AllIdle will be triggered when
+    *                       no read was performed for the specified period of time on given channel. Specify 0 to disable.
```
aakoshh: I didn't expect this: when I saw `AllIdle` I thought it means both read and write have gone idle, which would mean the timeout for it is `max(readerIdleTime, writerIdleTime)`, and if there's a write or a read than it goes back to just the other one being idle. 

BTW it says "no read was performed".

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
```diff
@@ -309,6 +310,53 @@ object DynamicTLSPeerGroup {
...
+      val writerIdleTime: Long,
+      val allIdleTime: Long
+  )
+
+  object StalePeerDetectionConfig {
+    def apply(
+        timeUnit: TimeUnit,
+        readerIdleTime: Long,
+        writerIdleTime: Long,
+        allIdleTime: Long
```
aakoshh: Why not use `FiniteDuration`? You can use whatever time unit you want inside, but the chance of misconfiguration would be lower, e.g. starting with `HOUR,1,1,1` then modifying it to `MINUTE,10,1,1` or something.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
```diff
@@ -69,11 +83,18 @@ private[peergroup] object DynamicTLSPeerGroupInternals {
...
+      eventLoop: EventLoop
   ) extends ChannelInboundHandlerAdapter
       with StrictLogging {
 
+    private def idleEventToChannelEvent(idleStateEvent: IdleStateEvent): ChannelIdle = {
+      idleStateEvent.state() match {
+        case IdleState.READER_IDLE => ChannelIdle(ReaderIdle, idleStateEvent.isFirst)
+        case IdleState.WRITER_IDLE => ChannelIdle(WriterIdle, idleStateEvent.isFirst)
+        case IdleState.ALL_IDLE => ChannelIdle(AllIdle, idleStateEvent.isFirst)
+      }
```
aakoshh, line 102: Some comments would be helpful on the `Channeldle` class about what to do with `isFirst` 
* When it goes `READER -> ALL -> READER`, is the second `READER` a first? 
* If a connection idles, then there's activity, then idles again, is the second one a first? 
* Are idle events raised just once per idleness? Or once per timeout, if it keeps being idle?

aakoshh (commented):
Great, more power to Scalanet :smiley: 

I'd just like to see some hints in our own docstrings to what expect, so we don't assume that everything is a mirror of Netty and everyone should go read that. Or if that's the case then the docs should be linked to.

I also think that we'd be well served with just 2 flags, and the "ALL" should be derived from them, or explained why it has to exist on its own, with a potentially different timeout then the other two.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: ahh it is an typo, it should read `neither read nor write was performed ` , so it is separate idle event idicating that nothing happened on this channel over this amount of time

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: I thought about it but ultimately this conveys intent that all things should have the same unit better, also using finite duration is error prone it way that user of library can pass `5.second` in one field, and `5.days` in other. 

One thing that is in favour of finite duration is that it is more scala friendly api, but I am not sure that is enough advantage.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
KonradStaniec: tbh I am not great fan of boolean flags in public interfaces, and three of them just make me feel the pain of writing more complex logic on top of it. Also `AllIdle` is totally separate event which can be  configured with totally different time out than other two events.  It maybe matter of opiniong, but imo
```
state match {
 ReaderState =>
 WriterState =>
 AllState =>
}
```
is easier on the eyes than:
```
if (raderIdle && writerIdler) {
} else if (readerIdler) {
} else {
}
```
especially if we were to add `first` flag to the mix.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroupInternals.scala
KonradStaniec: sure thing I will add those docs :+1:

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: Well you can pattern match on 
```scala
state match {
  case all @ ChannelIdle(true, true, _) =>
  case read @ ChannelIdle(true, _, _) =>
  case write @ ChannelIdle(_, true, _) =>
}
```
You an also make custom extractors as well:
```scala
object AllIdle {
  def unapply(idle: ChannelIdle): Boolean = 
    idle.readIdle && idle.writeIdle
}

state match {
  case AllIdle() if state.isFirst => ...
  ...
}
```

I guess my main problem is that there's no explanation to _why_ `AllIdle` is a "totally separate event with totally different timeout". Why isn't it just the case of read and write both being idle, according to their respective timeouts?

What would this mean for example? all = 1 minute, read = 5 minutes, write = 10 minutes? Or the opposite: read = 1 minute, write = 1 minute, all = 120 minutes?

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
aakoshh: `FiniteDuration` also works with the HOCON configuration. I think if the user configures 5 days, that's their problem, it's easy enough to read, and for us to find a common denominator.

aakoshh: `FiniteDuration` also works with the configuration. I think if the user configures 5 days, that's their problem, it's easy enough to read, and for us to find a common denominator.

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
KonradStaniec: so having some thing along the line :
` case all @ ChannelIdle(true, true, _)` looks pretty terrible imo with this amount of booleans, and custom extractors looks like work around for just having `AllIdleState`

As for _why_,  as I see it the main use case for `AllIdle` is when some does not care about read timeout and write timeout, and just want to say `if nothing happens on this channel for this amount of time get me this event` and it is pretty intuitive for this case to have `allIdleTimout` config option and corresponding `AllIdleState` event

As for your question, `read = 1 minute, write = 1 minute, all = 120 minutes` would mean that every minute you get `WriteIdle` and `ReadIdle` events , and every 120 minutes you get `AllIdle` events (assuming nothing happens). It is pretty convoluted but maybe you want to only log things for read and write, and if nothing happens for 120min you want to disconnect.

scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSPeerGroup.scala
KonradStaniec: I am still not fully convinced, but maybe having `FinititeDuration` for the sake of easy reading configs from HOCON seems neat. lets go this way then 👍

scalanet/src/io/iohk/scalanet/peergroup/PeerGroup.scala
aakoshh: The syntax can be worked around both ways, I wouldn't get hung up on that. It's more the mental model that's different.

For me the natural way to think about it was that reads and writes can time out, and `AllIdle` just means both being idle at the same time. 

But your example is enlightening: `ReadIdle` doesn't mean what I thought it did; it doesn't mean that "read is idle, write is not idle", and vice versa `WriteIdle` doesn't mean that reads happened, but writes didn't. These settings mean how frequently a notification event is supposed to be raised, which changes everything. 

In that case if I set `read = 1 minute, write = 5 minutes` then after 7 minutes I'd see `read; read; read; read; read, write, read; read`, so I'd never see a `readIdle=true, writeIdle=true` combination _unless_ I enable a timeout for `AllIdle`. But then I would have to disable read and write timeouts otherwise I'd get mixed signals, or I'd have to have some aggregator to make sense of it, which in itself would be complicated because it would have to listen to the _absence_ of notifications.

If you could an example which highlights the fact that we're talking about independent notification events (partial states) it would make it much clearer.

I agree it's convoluted.

aakoshh (approved):
Cool, thanks for changing the duration config! I think it's nice not to have to come up with an encoding for time units. And thank you for explaining the notification periods, makes sense now.

